PR #23, opened 2016-08-29
Docker readme quickstart create volumes

nwoetzel:
Trying the docker quickstart documented in docker/README.md
It is necessary to create the volumes in advance

> docker-compose version 1.8.0, build f3628c7
> Docker version 1.12.1, build 23cf638

I am not sure if there is a command line flag to docker-compose for the automatic creation of external volumes - which could be documented alternatively.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -7,6 +7,12 @@
 
 ## Quick start
 
+Create docker volumes:
+
+    docker volume create --name=seek-filestore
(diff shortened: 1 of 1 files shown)


stuzart:
Thanks.
Documenting docker is still in progress and needs transferring to the github pages, but this is a good short term update.

PR #25, opened 2016-09-06
use rails-assets for gems distributing javascript assets updating bootstrap plugins

nwoetzel:
bootstrap-multiselect; bootstrap-typeahead and bootstrap-tagsinput are in updated version now (using rails-assets.org)
the fix in the bootstrap-tagsinput-fix.js is now resolved differently (see the commit message)
added several ';' in app/assets/javascript/*.js so that the eclipse IDE warning list is not crowded by those warnings
bootstrap_tab_memory.js: showTabFromHistory did use attribute selector not quite compatible...


stuzart:
thanks.

PR #26, opened 2016-09-15
Publication form

nwoetzel:
added an alternative way to add a publication through an actual form
-in that form, doi and.or pubmedid can be entered and the form can be filled client-side
-authors are added in a tag-input field with some coloring depending on if seek has seen that author or even that user before
in the publication_controller:
-create function has a subaction:
  -'Register' is using the old way
  -'Create' is using the new form
-two action that assist in typeahead.js querying with bloodhound

Publication has changed:
-doi or pubmedid are not required any more and both can be given

typeahead.js upgrade from 0.10.4 -> 0.10.5

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -111,7 +111,7 @@ gem 'bundler', '>= 1.8.4'
 source 'https://rails-assets.org' do
   gem 'rails-assets-bootstrap-multiselect', '~> 0.9.13'
   gem 'rails-assets-bootstrap-tagsinput', '~> 0.8.0'
-  gem 'rails-assets-typeahead.js', '~> 0.10.4'
+  gem 'rails-assets-typeahead.js', '~> 0.10.5'
 end
 
 group :assets do
--- a/Gemfile
+++ b/Gemfile
@@ -135,6 +135,8 @@ group :development do
   gem 'guard-rubycritic', require: false
   gem 'rails_best_practices'
   gem 'quiet_assets'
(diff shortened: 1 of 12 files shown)


nwoetzel:
Could you please restart the build, it failed due to HTTP connection timeouts or not found errors in the bundle install step. Thanks

stuzart:
There may be a problem with this. It includes a major feature change in allowing publications with neither a DOI's or pubmedID. This was an early design decision to only allow published and citable publications. We need to discuss this more widely and it may need to be configurable. There was a recent decision to keep it as it was, although we have also had requests to allow pre-print publications.

nwoetzel:
I will close this request then and modify it by reverting that aspect of the code.
Actually, there are thoughts of extending publications beyond journal-articles (books, software, conference contributions without doi, pre-publications as you mentioned). For that, another agreement or setup will need to be found anyways.

stuzart:
No need to be hasty and close it. I don't think there is a particular problem, just needs checking with others first before merging, or a simple config option adding, or a way of clearly labelling it a pre-print etc. I can see there are clear benefits.

nwoetzel:
Sorry for closing it, but I will prepare another pull request that contains changes that need no previous discussions. I also wrote some tests, so that the next pull-request is a little more up to standard.
For the other mentioned aspects, I will need to get more precise requirements from the stakeholders at HITS - with those, we can start a discussion on how things should be changed or extended to suite the needs at HITS and anywhere else.

PR #29, opened 2016-10-04
Publication form calendar_date_select, https://dx.doi.org, add author on validation error

nwoetzel:
the publication date uses the calendar_date_select form helper 
filling the manual publication form using doi, https://dx.doi.org is used instead of http://data.crossref.org (no https available)
the tags-input for authors was not adding the tags for already entered authors correctly - the add expected an object with first_name and last_name but just got the full_name as a single string


nwoetzel:
Is there a way to restart coveralls for that pull request with the new commits - since there are new tests for all methods that have been added to the controller?

PR #41, opened 2016-11-15
LDAP simple authentification plugin

cmonjeau:
Hello,

First thank you for this tool because it answer to many questions about scientific collaboration in my team.

I've made a simple plugin to use LDAP authentification. 

This plugin add a section in the admin panel with some LDAP attributes (base DN , address, mail, uid, etc.) in order to match with several LDAP configurations. 

An LDAP user can directly log in with login/password and the register page appears with informations extract from the LDAP. I think it's very simple to configure for the admin (wordpress similar).

I'm sorry for the missing things or mistake but it's my first time with ruby-on-rails :). So I presume this plugin is not perfect.

Seekly,
Cyril


stuzart:
I'm afraid we recently had a contribution providing LDAP support through the Omniauth plugin: https://github.com/seek4science/seek/pull/31

We do like that your pull request includes the settings the admin pages with an AJAX test button.
Do you think it is possible the 2 could be combined? i.e with LDAP handled by the omniauth, but your support for configuration through the browser?

cmonjeau:
Hello stuzart,

during the dev of this plugin, I saw that a pull request was done about LDAP auth and I tried to use it. The fact is I have no cknowledge about omniauth and the configuration is not clear for me. So I continued to develop this plugin and that's why I call it "simple LDAP auth" ^^. 

I'm sorry but it was not easy to me to develop in ruby on rails and my cknowledge regarding omniauth are not sufficient to combine these two methods.

fbacall:
Hi Cyril,

If you want to try using omniauth for LDAP in SEEK, you can create a file in `config/initializers` called `seek_local.rb` with the following contents:

```
SEEK::Application.configure do
  Seek::Config.default :omniauth_enabled, true
  Seek::Config.default :omniauth_user_create, true
  Seek::Config.default :omniauth_user_activate, true
  Seek::Config.default :omniauth_providers, {
    # https://github.com/intridea/omniauth-ldap
    :ldap => {
      :title => "organization-ldap",
      :host => 'localhost',
      :port => 389,
      :method => :plain,
      :base => 'DC=example,DC=com',
      :uid => 'samaccountname',
      :password => '',
      :bind_dn => ''
    }
}
end
```

cmonjeau:
Hi @fbacall,

thanks for this information. 

I see that the configuration is inside the code so it's difficult to edit and test. 

In addition, my LDAP simple plugin fill some fields automatically so the user hasn't the impression to create a new account.

Actually  I'm playing with docker and your image (it's a great work, thanks) and the LDAP config in front-end is very useful because I use the openldap docker image.

So I prefer my LDAP plugin for my uses :)...but I understand if you want to reject this pull request.

Maybe you can use this work to complete the existing plugin?

Thanks,
Cyril

fbacall:
Yes, you are right that it is not possible to set through the interface.

I think our Omniauth/LDAP configuration should also fill in fields automatically:
https://github.com/seek4science/seek/blob/master/app/controllers/sessions_controller.rb#L143
(I haven't tested this because I didn't write the code and don't have an LDAP back-end to test with)

I think this PR adds useful features such as the LDAP config in the admin interface and the testing tool, so it would be good if we could integrate that with our existing LDAP solution.

cmonjeau:
Hi all,

I've tried the omniauth plugin and it works but: 

- when a username/password is wrong, there is a 404 error page instead of a message.
- there are two frames, one for SEEK users and the other for LDAP. Like I've only LDAP users, it's pretty troublesome in my use.
- When a LDAP user connects for the first time, there are not sign up page so the user can't join projects or ask new projects / institutions.

That's all for my tests. It's a very different plugin afterall ^^.
Cheers

PR #42, opened 2016-11-17
Add an option to disable users registration

cmonjeau:
Hello,

 I think it's important to disable the public registration. For example when we want limit the number of users or when we use an external authentification (e.g. LDAP).

So I've added an option in the admin panel in order to disable the registration. This plugin replace the registration form by a message. This message can be modify in the option.


stuzart:
Looks good, thanks !

stuzart:
I have noticed, and a possible refinement, but there is not server side check for somebody registering with it turned off. It could be possible for somebody to register using a script (such as using curl to post the params), if they are that way inclined, and get around the disabled registration.

cmonjeau:
I agree with you. This option is more to prevent users registration than hackers registration ^^.

Actually, I use this feature with my simple LDAP auth where there is a registration on server side when a new user log-in ; so I've preferred not to make changes.

PR #43, opened 2016-11-22
bugfix: footer does not cover part of the content anymore in IE

susca:
Before, lower parts of the content were not shown in internet explorer (tested with IE11). That made some forms (e.g. "Edit Profile") unusable, because the update and cancel buttons were not visible. All other browsers worked as expected.

The reason was that the footer covered parts of the content in IE.

The solution is to get rid of the "bottom: 0px;" style for id "#ft".


stuzart:
thanks!

PR #56, opened 2017-09-15
Handling custom properties of OpenBIS Dataset

tzielins:
Few tweaks to make the dataset properties visible and searchable in SEEK. 
Most of the code was already there I only needed to add rendering in the view and deal with indexing and reindexing.


tzielins:
will submit reformatted code for the pull

PR #58, opened 2017-10-20
Update broken methodbox.org link

stain:


fbacall:
The link checker says some of the links are broken but they work for me..

PR #65, opened 2018-02-21
Link DOIs to preferred resolver

katrinleinweber:
related to #29

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). This PR implements that recommendation by updating static DOI links, as well as code that generates new ones.

Since I updated some test code as well, please let me know if that depends on external services who still answer with the "old" resolver URL, and I'll revert that part.

Cheers :-)

### DIFF
--- a/test/unit/publication_test.rb
+++ b/test/unit/publication_test.rb
@@ -415,11 +415,11 @@ class PublicationTest < ActiveSupport::TestCase
     assert pub.valid?
     assert_equal '10.5072/abc', pub.doi
 
-    pub.doi = 'http://dx.doi.org/10.5072/abc'
+    pub.doi = 'https://doi.org/10.5072/abc'
     assert pub.valid?
     assert_equal '10.5072/abc', pub.doi
 
-    pub.doi = 'https://dx.doi.org/10.5072/abc'
+    pub.doi = 'https://doi.org/10.5072/abc'
     assert pub.valid?
     assert_equal '10.5072/abc', pub.doi
 
--- a/test/unit/publication_test.rb
+++ b/test/unit/publication_test.rb
@@ -419,7 +419,15 @@ class PublicationTest < ActiveSupport::TestCase
     assert pub.valid?
     assert_equal '10.5072/abc', pub.doi
 
-    pub.doi = 'https://doi.org/10.5072/abc'
+    pub.doi = 'https://dx.doi.org/10.5072/abc'
+    assert pub.valid?
+    assert_equal '10.5072/abc', pub.doi
+
+    pub.doi = 'http://dx.doi.org/10.5072/abc'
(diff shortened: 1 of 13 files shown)


fbacall:
Hi Katrin, thanks for this.

Looks like some of the tests are failing, and I think it's because you need to change the `doi=` method in the publication model to substitute the new DOI resolver URL. Although I think it would be good if this method recognized both the old and new resolver URLs.

test/unit/publication_test.rb
```diff
@@ -415,11 +415,11 @@ class PublicationTest < ActiveSupport::TestCase
     assert pub.valid?
     assert_equal '10.5072/abc', pub.doi
 
-    pub.doi = 'http://dx.doi.org/10.5072/abc'
+    pub.doi = 'https://doi.org/10.5072/abc'
     assert pub.valid?
```
fbacall: Could this test check both the old and new resolver URLs?

katrinleinweber: Phew... Testing in Ruby is a bit out of my waters, so I tried duplicating the input strings & asserts for the other URL variants, see below.

katrinleinweber:
> […] change the doi= method in the publication model to [make] recognize[] both the old and new resolver URLs.

Like in d6316ae?

fbacall:
Yeah, that looks good 👍 , lets see what travis says...

katrinleinweber:
Meh, already `Failure`:

```
PublicationTest#test_strips_domain_from_DOI_if_an_URL_is_given [/home/travis/build/seek4science/seek/test/unit/publication_test.rb:419]
Minitest::Assertion: Expected false to be truthy.
```

fbacall:
Right, sorry, I didn't look at your commit properly...
You'll need to modify this setter method in the model:
https://github.com/katrinleinweber/seek/blob/3d198a6e23560d74a8b951efdfe5d9f4b4d57547/app/models/publication.rb#L88

katrinleinweber:
OK, thanks for the hint :-) Now it is a bit different from https://github.com/katrinleinweber/seek/blob/resolve-DOIs-securely/app/views/publications/new.html.erb#L314. Shall those two reg-exes be the same? Specifically, both including `(https?:\/\/)?`?

fbacall:
Yeah, I think the way it is done in the model is more robust, with the optional `(https?:\/\/)?`.

fbacall:
Looks like there are still a few old "dx.doi..." references hanging around in regular expressions.

If you do a global search/grep for `dx\.doi` you should get the last of them (hopefully)

stuzart:
Thanks !

katrinleinweber:
You're welcome :-) I thank you for this introduction to REs in Ruby!

PR #69, opened 2018-04-30
Openbis int merge

tzielins:
Implemented Experiment=>Study, Sample=>Assay, Datasets=>DataFile. This pull request is primary to check it with seek CI


tzielins:
Will re-submitted with fixes

PR #85, opened 2019-07-24
added content splitter

vdkkia:
The content is split into 10 word with an overlap of 5 words (by default)

### DIFF
--- a/lib/seek/content_split.rb
+++ b/lib/seek/content_split.rb
@@ -1,0 +1,17 @@
+module Seek
+  module ContentSplit
+    def split_content(str, length = 10, overlap = 5)
+      overlap = length - overlap
+      raise 'overlap should be smaller than length' if overlap <= 0
+      length -= 1
+      words = str.split(' ')
+      phrases = []
+      loop do
+        temp = words[phrases.length * overlap..phrases.length * overlap + length]
+        break if !temp || temp.empty?
+        phrases << temp.join(' ')
+      end
+      phrases
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/seek/content_split.rb
```diff
@@ -0,0 +1,17 @@
+module Seek
+  module ContentSplit
+    def split_content(str, length = 10, overlap = 5)
```
fbacall: any reason for choosing 10 and 5 as default parameters?

lib/seek/content_split.rb
```diff
@@ -0,0 +1,17 @@
...
+      overlap = length - overlap
+      raise 'overlap should be smaller than length' if overlap <= 0
+      length -= 1
+      words = str.split(' ')
+      phrases = []
+      loop do
+        temp = words[phrases.length * overlap..phrases.length * overlap + length]
```
fbacall: maybe add some parentheses to each side of the `..`

test/unit/content_split_test.rb
```diff
@@ -0,0 +1,18 @@
+require 'test_helper'
+class SplitTest < ActiveSupport::TestCase
+  include Seek::ContentSplit
+  test 'return instance of array' do
+    assert_instance_of  Array, split_content('Hello world!')
+  end
+  test 'extract content' do
+    assert_equal split_content('',10,5),[]
+    assert_equal split_content('Hello world!',10,5),['Hello world!']
+    assert_equal split_content('Hello world!',10,5),['Hello world!']
```
fbacall: perhaps add a test with some longer, more complex text

lib/seek/content_split.rb
vdkkia: There is really no solid reason for this other than ease of use, I will remove the defaults

stuzart:
this was merged manually a while ago, thanks.

PR #212, opened 2020-04-08
WIP: Add galaxy2cwl to Dockerfile

stain:
Adds `galaxy2cwl` on path, but is not yet called by the galaxy parser in Ruby


stain:
I also added `cwltool`. Perhaps *TODO* is to move the `pip install` arguments out to a `requirements.txt` that can be installed also outside Docker?

fbacall:
> I also added `cwltool`. Perhaps _TODO_ is to move the `pip install` arguments out to a `requirements.txt` that can be installed also outside Docker?
:+1: to this. We have quite a few non-container deployments

stuzart:
it should also go in the install docs and the upgrade notes

stuzart:
I'll be adding the requirements.txt as part of getting the travis working correctly, it seems to be the recommended way

Dockerfile
```diff
@@ -13,18 +13,23 @@ RUN apt-get update -qq && \
...
 RUN mkdir -p $APP_DIR
-RUN chown www-data $APP_DIR
+RUN chown -R www-data $APP_DIR /var/www
 
 USER www-data
 
 WORKDIR $APP_DIR
 
+# CWL dependencies, galaxy2cwl
+RUN pip3 install cwltool==3.0.20200324120055 html5lib==1.0.1 galaxy2cwl==0.1.4
```
stuzart, line 30: this could use the requirements.txt which contains cwltool, html5lib and galaxy2cwl

PR #239, opened 2020-05-08
feature/disease-ontology-integration

ChristophB:
[Human Disease Ontology](https://disease-ontology.org) integration from the [Leipzig Health Atlas](https://health-atlas.de).

We made the following changes:

* new toggleable feature human diseases
* human diseases can be fetched from BioPortal (with all super concepts)
* new index page for human diseases, which contains a jsTree
* new navigation entry links to the human disease index and create pages
* projects, publications, assays and models can be annotated with concepts of the human disease ontology
* annotations are visible on the respective show pages, in search results and in the newly added filter of SEEK v1.10
* new section on the front page with sunburst diagrams to the current state of tagged entities (diagram data are cached for 12 hours)
* new tests related to human diseases

Originally these changes were made on v1.9.1, that's why I merged *master* into this branch.


stuzart:
Thanks, I've now merged the pull request into the main master branch. There was a slight delay as I needed to wait for other branches to be merged across first. 
The merge went smoothly, I just needed to add the enable configuration to the element on the front page.
Many thanks for this, there are components that I feel we can reuse for work around ontology annotations we are now starting to move on to.

PR #455, opened 2020-11-03
WIP: Study batch with metadata

Jerem-info:
Add a study batch upload with MIAPPE metadata, to create studies from a MIAPPE standardized file.


config/locales/en.yml-openseek
```diff
@@ -11,6 +11,7 @@ en:
 
   investigation: "Investigation"
   study: "Study"
+  studies: "Studies"
```
stuzart, line 14: this shouldn't be needed - you can do t('study').pluralize

app/models/sample_attribute.rb
```diff
@@ -32,6 +32,10 @@ def required?
     super || is_title?
   end
 
+  def pre_process_value(value)
+    sample_attribute_type.pre_process_value(value, controlled_vocab: sample_controlled_vocab, linked_sample_type: linked_sample_type)
+  end
+
```
stuzart, line 38: is this method needed? I was recently removed thinking it was now unused - I wonder if it reappeared due to a merge problem

stuzart:
thanks

Issue #392, opened 2020-07-16
Visibility of old versions


fbacall:
Currently only the latest version of a resource is visible to anonymous users.

It was agreed to change it so that going forwards, new versions are visible to anyone. Existing, old (not latest) versions should still be hidden unless logged in.

Versions with an associated DOI should also be visible to anyone.

stuzart:
we still want the option to hide versions for new items as well

PR #467, opened 2020-11-19
Allow versions to be hidden

fbacall:
Add a flag to version tables to say if the version is hidden/visible/visible only to registered users. 

Allows users with `manage` rights to change the visibility of a version, unless it is the latest version, or has an associated DOI.

Related issue: #392


stuzart:
all looks good to me. I checked in particular there was an upgrade task, and it wasn't possible to hide the latest version

Issue #426, opened 2020-09-07
Finishing off the registration flow and joining a project


stuzart:
Some tasks are remaining to complete the work, that was completed only as a prototype for the CoFest for feedback.

- [x] Sign up email needs updating to hide the project & institution details
- [x] Handle programmes disabled better
- [x] Banner and page for admins to show there are outstanding requests
- [x] Handle search disabled
- [x] Handle email disabled
- [x] Reuse for general project creation
- [x] Reuse for project creation for a programme
- [x] Handle no managed programme
- [x] Handle user creation of programmes disabled
- [x] Reuse the institution finder for adding project members
- [x] Setting the system administered programme in the UI
- [ ] Better use of nested attributes
- [x] General code tidying up
- [x] Extended testing
- [x] Notification of pending creation requests in the UI, and page to handle
- [x] Notification of pending join requests in the UI, and page to handle
- [x] Tweak the text if you are a site or programme admin

PR #505, opened 2021-02-01
Ensure Gender neutral They instead of he/she in email template

stain:
Some of the email templates still used "he/she" to talk about a user in third person. Changed to gender-neutral "they" to avoid assumption of binary gender.


stain:
I checked the other email notifications in https://github.com/seek4science/seek/tree/gender-neutral/app/views/mailer which all are either in 2nd person ("you") or use gender-neutral "they"

stain:
I also got this email from WorkflowHub, which template text I could not find:

> Hi Admin,
>
> The new user ----- just signed up. He/She left the following details about their Team and Organization affiliations:
> There are NO Team Administrators for these Teams.

I think that template was already fixed by #426

stuzart:
thanks @stain . Yes, the other template was removed or replaced with #426

PR #524, opened 2021-02-26
Rocrate jsonld

alaninmcr:


lib/seek/bio_schema/resource_decorators/base_decorator.rb
```diff
@@ -26,7 +26,7 @@ def attributes
 
         # The @context to be used for the JSON-LD
         def context
-          'http://schema.org'
+          'https://schema.org'
```
fbacall, line 29: Not sure about this...

The schema.org site uses both http and https in examples, but the RO-Crate context uses http. Maybe @stain can comment

test/unit/workflow_test.rb
```diff
@@ -57,7 +57,7 @@ class WorkflowTest < ActiveSupport::TestCase
   end
 
   test 'generates fresh RO crate for individual workflow' do
-    workflow = Factory(:cwl_workflow, license: 'MIT', other_creators: 'Jane Smith, John Smith')
+    workflow = Factory(:cwl_workflow, license: 'https://opensource.org/licenses/MIT', other_creators: 'Jane Smith, John Smith')
```
stuzart, line 60: This looks wrong. The licenses are stored as key which then then turned into a URL as part of the schema.org generation. They shouldn't be directly stored as a URL

app/models/concerns/workflow_extraction.rb
```diff
@@ -80,22 +80,59 @@ def populate_ro_crate(crate)
...
     crate.preview.template = PREVIEW_TEMPLATE
+    # brute force deletion as I cannot track down where it comes from
+    crate.contextual_entities.delete_if { |c| c['@id'] == '#ro-crate-preview.html' }
   end
 
+  def merge_entities (crate, workflow)
+    workflow_struct = Seek::BioSchema::Serializer.new(workflow).json_representation
+
+    context = {
+      '@vocab' => 'https://somewhere.com/'
```
fbacall, line 109: what is this for?

app/models/concerns/workflow_extraction.rb
```diff
@@ -80,22 +80,59 @@ def populate_ro_crate(crate)
...
+    }
+    workflow_struct['@context'] = context
+    workflow_struct.except!('encodingFormat')
+
+    flattened = JSON::LD::API.flatten(workflow_struct, context)
```
fbacall, line 114: how long does this take?


In my experience, using any kind of linked-data aware library takes 10+ seconds to do any operation, and in SEEK we can check a resource's RO-Crate several times in a request.

test/functional/workflows_controller_test.rb
```diff
@@ -416,7 +416,7 @@ def setup
       post :create_ro_crate, params: {
           ro_crate: {
               workflow: { data: fixture_file_upload('files/checksums.txt') },
-              diagram: { data: fixture_file_upload('files/file_picture.png') },
+              #             diagram: { data: fixture_file_upload('files/file_picture.png') },
```
fbacall, line 419: Is there a reason for these being commented out?

app/models/concerns/workflow_extraction.rb
```diff
@@ -80,22 +80,59 @@ def populate_ro_crate(crate)
...
+#    others = other_creators&.split(',')&.collect(&:strip)&.compact || []
+#    authors += others.map.with_index { |name, i| crate.add_person("creator-#{i + 1}", name: name) }
+#    crate.author = authors
+#    crate['provider'] = projects.map { |project| crate.add_organization(nil, project.ro_crate_metadata).reference }
+#    crate.license = license
+#    crate['sdPublisher'] = crate.add_person(nil, contributor.ro_crate_metadata).reference
 
     crate.preview.template = PREVIEW_TEMPLATE
+    # brute force deletion as I cannot track down where it comes from
+    crate.contextual_entities.delete_if { |c| c['@id'] == '#ro-crate-preview.html' }
```
fbacall, line 102: This is fixed in a recent release of the RO-Crate library

lib/seek/bio_schema/resource_decorators/creative_work.rb
```diff
@@ -25,7 +25,7 @@ def license
 
         def all_creators
           others = other_creators&.split(',')&.collect(&:strip)&.compact || []
-          others = others.collect { |name| { "@type": 'Person', "name": name } }
+          others = others.collect { |name| { "@type": 'Person', "@id": name.downcase.gsub(/[^0-9a-z]/i, '_'), "name": name } }
```
fbacall, line 28: I think you should use proper URI-encoding here, since its fairly common to have names with non-ASCII characters. Also the ID should be prefixed with `#`

test/unit/workflow_test.rb
alaninmcr: Does this mean we can change them to point to spdx?

lib/seek/bio_schema/resource_decorators/workflow.rb
```diff
@@ -22,28 +24,38 @@ def image
...
+          resource.license
         end
 
         private
 
-        def formal_parameters(properties)
+        def formal_parameters(properties, group_name)
+          if self.title
+            wf_name = self.title.downcase.gsub(/[^0-9a-z]/i, '_')
+          else
```
fbacall, line 51: Is this needed? title is mandatory

test/unit/workflow_test.rb
stuzart: yes, we already could. https://github.com/seek4science/seek/blob/seek-1.11/lib/seek/bio_schema/resource_decorators/creative_work.rb#L23
or possibly use identifier.org

stuzart: if this is to get the url for the ro-crate, then it would be better to make the change there rather than changing the behaviour of Workflow to be different from other assets. You could add a licence_url to all licensed assets

lib/seek/bio_schema/resource_decorators/base_decorator.rb
stuzart: I was unsure about this too. https didn't used to validate using the structured data tool, and the schema.org examples are quite mixed up. I double checked and https now does seem to validate

alaninmcr: Yes. The choice confused me too. The advice from schema.org doesn't really help

There is a general trend towards using 'https' more widely, and you can already write 'https://schema.org' in your structured data. Over time we will migrate the schema.org site itself towards using https: as the default version of the site and our preferred form in examples. However 'http://schema.org' -based URLs in structured data markup will remain widely understood for the forseeable future and there should be no urgency about migrating existing data. This is a lengthy way of saying that both 'https://schema.org' and 'http://schema.org' are fine

app/models/concerns/workflow_extraction.rb
alaninmcr: It is a hacky way of making darned sure that the aliasing of @id -> id does not happen

app/models/concerns/workflow_extraction.rb
alaninmcr: It could be hand coded I guess. Would that be better?

test/functional/workflows_controller_test.rb
alaninmcr: That is what was causing the diagram route complaints that I never sorted out

lib/seek/bio_schema/resource_decorators/creative_work.rb
alaninmcr: OK.

lib/seek/bio_schema/resource_decorators/base_decorator.rb
stuzart: perhaps we just need to be consistent with bioschemas.org and ro-crate and migrate together

lib/seek/bio_schema/resource_decorators/workflow.rb
alaninmcr: Then possibly not,

test/unit/workflow_test.rb
alaninmcr: I'm not clear what you are suggesting.

stuzart: OK. 
You seem to have overridden the behaviour for Workflow and then manipulated the tests to store the full URL, but in reality it would still be the key. 
This seems to be so that the ro-crate generated would have a license URL by reading workflow.license but I think this will only work for the test, due to the full URL being directly stored.
You would need to call `Seek::License.find(workflow.license)&.url` for the ro-crate too, but rather than duplicating this for both ro-crate and the schema.org you could create a new `license_url ` method that does this lookup in one place.

fbacall: https://github.com/seek4science/seek/blob/workflow/lib/seek/workflow_extractors/ro_crate.rb#L61

would also need to be changed to convert an incoming license URL to a license ID that SEEK recognizes

stuzart: so is `workflow.license` currently being stored as either a URL or ID depending if it was submitted as an RO-Crate or not ? Entries I've looked at through the JSON API use the ID, e.g https://workflowhub.eu/workflows/52.json

fbacall: it's always an ID at the moment, but this is not valid according to bioschemas or schema.org

I mention it here: https://about.workflowhub.eu/Workflow-RO-Crate/#supported-licenses

stuzart: yes, in the bioschema.org markup generated for the page it was getting translated to a URL from the stored ID. e.g


```<script type="application/ld+json">{
"@context": "http://schema.org",
"@type": "ComputationalWorkflow",
"@id": "https://workflowhub.eu/workflows/52",
...
"license": "https://opensource.org/licenses/MIT",
...
```
it just needs to do the same for the RO-Crate to get the same URL and be valid

as long as the ID's at https://about.workflowhub.eu/Workflow-RO-Crate/#supported-licenses are the same as opendefinition and spdx which they seem to be, it should be ok.
It is possible to change the URL conversion to create SPDX or Indentifiers.org URLs for software licenses in the future quite easily.

@alaninmcr's change will mean the invalid ID will be shown in both the page markup and, from what I understand, also the RO Crate

stuzart: https://about.workflowhub.eu/Workflow-RO-Crate/#supported-licenses could be changed to list of recommended SPDX urls instead, and they could be easily translated back to ID's seek recognises.

lib/seek/bio_schema/resource_decorators/creative_work.rb
fbacall: You can do this using the ROCrate library: `"##{ROCrate::Entity.format_id(name)}"`

app/models/concerns/workflow_extraction.rb
fbacall: possibly, and would make the above hack unnecessary

test/unit/workflow_test.rb
alaninmcr: OK. The spdx suggest that the SPDX IDs should be used whereever possible. There is a JSON file that contains the list of spdx licenses https://github.com/librariesio/spdx/blob/master/licenses.json

I think we should leverage that file.,

alaninmcr: Just for this problem, I have changed to use the constant e.g. 'MIT'. I have not made the change to spdx

lib/seek/bio_schema/resource_decorators/base_decorator.rb
alaninmcr: http://schema.org is forwarding to https://schema.org so that is the one we should use

app/models/concerns/workflow_extraction.rb
alaninmcr: For an initial version - I do not think this is necessary

app/models/concerns/workflow_extraction.rb
alaninmcr: It is still coming through. I updated to ro_crate 0.4.9 and it still comes through. So I am leaving these two lines in.

lib/seek/bio_schema/resource_decorators/creative_work.rb
alaninmcr: Done

lib/seek/bio_schema/resource_decorators/workflow.rb
alaninmcr: It does no harm

test/functional/workflows_controller_test.rb
alaninmcr: Uncommented. Now checking if it builds

alaninmcr: It didn't and I cannot sensibly debug.

stain:
Do we just need to resolve the conflicts before merging this?

PR #580, opened 2021-04-28
Ordering

alaninmcr:
Ordering for review


app/views/assays/show.html.erb
```diff
@@ -68,7 +72,14 @@
     <% if @assay.external_asset %>
       <div class="row">
         <div class="col-md-12">
-          <%= external_asset_details(@assay) %>
+          <%= external_asset_details(@assay) %>tudy.all.each do |s|
```
stuzart, line 75: I spotted this strange bit of code whilst resolving conflicts against another branch. Looks like added by error so I stripped it out (a similar case below too).

alaninmcr: Yes mistake

Issue #559, opened 2021-04-01
Omniauth LDAP : invalid credentials redirects to 404


thopico:
**Describe the bug**
When entering invalid credentials with LDAP we face a 404 not found page (see screenshot 1).
The hit is `auth/failure?message=invalid_credentials&strategy=ldap`.

**To Reproduce**
1. Go to 'Login'
2. Click on 'LDAP login'
3. Enter wrong credentials

**Expected behavior**
An error message such as the one used with SEEK login (see screenshot 2).

**Screenshots**
1.
![Capture d’écran de 2021-04-01 12-27-07](https://user-images.githubusercontent.com/56619719/113281360-b8c4bf80-92e5-11eb-9cad-e0cf741f15ef.png)
2.
![Capture d’écran de 2021-04-01 12-27-45](https://user-images.githubusercontent.com/56619719/113281362-b9f5ec80-92e5-11eb-962e-d4f238d48cdb.png)

**Server (please complete the following information):**
 - Dockerized? yes
 - Version 1.10.3

Thank you for your help.

Issue #558, opened 2021-04-01
Possibility to hide SEEK login form when using only LDAP


thopico:
Hi, this is not a bug but a feature request.

**Describe the issue**
Our organization is only using LDAP as main authentication mechanism, and we don't want our users to use the « SEEK login » form to log in. In order to make it less confusing, we would like to have the possibility to hide the classic « SEEK login » form.

**Expected solution**
Have a setting which when set to false hides the « SEEK login » form.

**Screenshots**
![2021-04-01_11-01](https://user-images.githubusercontent.com/56619719/113270398-b78d9580-92d9-11eb-8dfc-fdd283174b0e.png)

Thank you for your time.

Linus4:
We're using the normal SEEK login for administration and LDAP for our users, so we need both. It would be nice if we could make LDAP login the default option because many users don't know about this and try to authenticate using the normal login.

![Screenshot 2022-04-26 at 13-42-20 The SEEK](https://user-images.githubusercontent.com/20419033/165292591-471d2fc6-b305-4808-b9ca-d95e7eb42942.png)

thopico:
Eventually we went after the same configuration and we managed to solve it with this change :
```diff
diff --git a/app/views/gadgets/_sign_in.html.erb b/app/views/gadgets/_sign_in.html.erb
index 2a704bbae4..9f4ada5145 100644
--- a/app/views/gadgets/_sign_in.html.erb
+++ b/app/views/gadgets/_sign_in.html.erb
@@ -1,5 +1,5 @@
 <% original_path ||= request.original_fullpath %>
-<% strategy = params[:strategy] || 'password' %>
+<% strategy = params[:strategy] || 'ldap' %>
 
 <% unless logged_in_and_registered? -%>
     <div class="panel panel-default" id="login-panel">
```

I'm not sure this custom feature should be generalized as is (because normal Seek login should stay the norm).
The perfect solution would be to have a button in the admin panel to select the preferred login mode, but I understand if it's too much detail for now.

So for now we are good with this manual change for our organization.
I will close the issue but feel free to reopen it if you want to address this 'login mode button' matter.

PR #587, opened 2021-05-04
Handle Ominauth auth failure

abretaud:
Hi,
This is a fix for #559 : added a route and action to properly handle authentication failure when using omniauth.
Instead of a 404 error, you get an error like that:

image

There is no "forgotten your password?" link in this case as we can't guess how the user is supposed to get its forgotten password from the external auth provider.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -44,6 +44,10 @@ class SessionsController < ApplicationController
     password_authentication
   end
 
+  def failure
+    failed_login "Invalid username/password.".html_safe
+  end
+
   def destroy
     logout_user
     flash[:notice] = 'You have been logged out.'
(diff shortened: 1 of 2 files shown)


fbacall:
Thanks for this. Does it automatically focus the correct tab? i.e. if it was an LDAP authentication failure does it show the LDAP tab?

abretaud:
Hum, you're right, the tab is not focused. I didn't see it because we (with @thopico) patched our instance locally to focus the ldap tab by default (because of #558).
I'm not a ruby expert and I think it's a bit more complex to know the login method that was used when reaching /auth/failure, do you see a way to solve this problem?

fbacall:
I have a few ideas, for example from the linked issue it looks like there is a `strategy` parameter passed with the value `ldap`, which could be checked.

Either way, I think this PR is an improvement to the current state where you just get a 404

PR #621, opened 2021-05-25
Multi sample linking

vdkkia:


app/controllers/samples_controller.rb
```diff
@@ -126,13 +126,33 @@ def filter
...
 
   def sample_params(sample_type=nil)
     sample_type_param_keys = sample_type ? sample_type.sample_attributes.map(&:title).collect(&:to_sym) : []
     if params[:sample][:attribute_map]
         params[:sample][:data] = params[:sample].delete(:attribute_map)
     end
+    sample_type_param_keys.each_with_index do |p,i|
+      if params[:sample][:data] && params[:sample][:data][p]
+        par = params[:sample][:data][p]
+        sample_type_param_keys[i] = {"#{p}":[]} if par && par.kind_of?(Array)
```
fbacall, line 153: I think this should not be checking the incoming parameters, but instead should check the `sample_attribute` is a `SeekSampleMulti`.

app/models/sample.rb
```diff
@@ -77,7 +77,11 @@ def referenced_resources
     sample_type.sample_attributes.select(&:seek_resource?).map do |sa|
       value = get_attribute_value(sa)
       type = sa.sample_attribute_type.base_type_handler.type
-      type.constantize.find_by_id(value['id']) if value && type
+      if value.kind_of?(Array)
+        value.map {|v| type.constantize.find_by_id(v['id']) if v && type} 
+      else
+        type.constantize.find_by_id(value['id']) if value && type
+      end
```
fbacall, line 84: Similar to above comment, you can force the value to an array (with `Array(v)`) to avoid duplication:
```
Array(value).map {|v| type.constantize.find_by_id(v['id']) if v && type} 
```

app/models/sample.rb
```diff
@@ -77,7 +77,11 @@ def referenced_resources
...
     end.compact
```
fbacall, line 85: You will also probably need to `.flatten` (before `.compact`) the array here, since it will contain nested arrays.

app/helpers/samples_helper.rb
```diff
@@ -62,6 +64,15 @@ def controlled_vocab_form_field(attribute, element_name, value)
     end
   end
 
+  def sample_multi_form_field(attribute, element_name, value)  
+    existing_objects = []
+    value.each {|v| existing_objects << Struct.new(:id, :name).new(v[:id], v[:title]) if v} if value
```
fbacall, line 69: `existing_objects` can also be a JSON string, so you can just pass `value.to_json` instead of converting to Structs

app/views/samples/_form.html.erb
```diff
@@ -25,5 +25,20 @@
...
+          const valuesArray = $j(item).val().split(",");
+          $j.each(valuesArray, (j,value) => {
+            $j(this).append(`<input type='hidden' name='${$j(item).attr('name')}[]' value='${value}'>`)
+          })
+          $j(item).siblings('input:hidden:first').val(JSON.stringify(valuesArray))
+          $j(item).attr('name', '')
+        })
+      });
+     
+  </script>
```
fbacall, line 42: Is this needed?

The `objects_input` in the form should provide the params as a comma separated list of IDs, that get handled by the `convert` method in the attribute type handler.

app/controllers/samples_controller.rb
fbacall: Actually I think it should recieve a comma separated String, so this extra code might not be needed at all.

lib/seek/samples/attribute_type_handlers/seek_sample_multi_attribute_type_handler.rb
```diff
@@ -0,0 +1,60 @@
...
+          else
+            test_value_item value
+          end
+        end
+
+        def convert(value)
+          if value.kind_of?(Array)
+            value.uniq.map {|v| get_conversion v}
+          else
+            if value.include? ','
```
fbacall, line 23: You don't need this check, since if you call `split` on a string that does not contain the value, it just returns itself in an array.

Can also reduce duplication by first making sure `value` is an array, then calling `map` etc.

lib/seek/samples/attribute_type_handlers/seek_sample_multi_attribute_type_handler.rb
```diff
@@ -0,0 +1,60 @@
+module Seek
+  module Samples
+    module AttributeTypeHandlers
+      class SeekSampleMultiAttributeTypeHandler < SeekResourceAttributeTypeHandler
```
fbacall, line 4: You could make this a subclass of `SeekSampleAttributeTypeHandler`, then for a lot of methods you can just split the array and call `super` to re-use the existing code to lookup the `Sample` object etc.

e.g.
```ruby
def convert(value)
  value = value.split(',') if value.is_a?(String)
  value.uniq.map { |v| super(v) }
end
```

app/helpers/samples_helper.rb
vdkkia: I think since the value has **title** key rather than **name**, I can't pass the value.to_json

vdkkia:
All the refactors except one in the unresolved conversation have been done.

app/helpers/samples_helper.rb
fbacall: I think in that case you can at least store the struct `Struct.new(:id, :name)` in a variable outside the loop, instead of making a new one for each iteration.

PR #647, opened 2021-06-25
Debian 10 dockerfile

### LINKED ISSUE #646: cwltool 3.0.20201203173111 not compatible with Docker container 20201203173111
7f9ae90 bumped the cwltool version in `requirements.txt` to `3.0.20201203173111` - however as the Debian 9 base image of `Dockerfile` (via `ruby:2.6-stretch`) is based on Python 3.5, building it will fail with:

```
Collecting cwltool==3.0.20201203173111 (from -r requirements.txt (line 1))
  Could not find a version that satisfies the requirement cwltool==3.0.20201203173111 (from -r requirements.txt (line 1)) (from versions: 0.1.20141118163626.cbb1b06, 0.1.20141118185349.d594c3e, 0.1.20141119031726.8dcd753, 0.1.20141125184103.2d72b5c
```

The last known Python 3.5 compatible `cwltool` is `3.0.20200807132242`.

Python 3.5 was deprecated with 5 Sept 2020 which is around the same time.

Suggestion: Change `Dockerfile` with `sed` to force use of older `cwltool==3.0.20200807132242` – or update  to `FROM ruby:2.6-buster` ?

stain:
Use Debian 10 base image ruby:2.6-buster
    
As this has Python 3.7 instead of 3.5 this should fix #646 –  however more testing might be needed as other dependencies also change version. 

So I am making this WIP pull requests to see if the tests work https://github.com/seek4science/seek/runs/2915756586

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.6-stretch
+FROM ruby:2.6-buster
 
 MAINTAINER Stuart Owen <orcid.org/0000-0003-2130-0865>, Finn Bacall
 
@@ -44,9 +44,6 @@ RUN chown -R www-data solr config docker public /var/www db/schema.rb
 USER www-data
 RUN touch config/using-docker #allows us to see within SEEK we are running in a container
 
-# Workaround for deprecated Python 3.5 from Debian base image, 
-# latest compatible cwltool version is from 2020-08
-RUN sed -i s/cwltool.*/cwltool==3.0.20200807132242/ requirements.txt
 # Python dependencies from requirements.txt
 RUN pip3 install -r requirements.txt
 


stain:
Never mind tests - as https://github.com/seek4science/seek/blob/workflowhub/.github/workflows/tests.yml never test the `Dockerfile` - separate bug needed to cover that in tests.

stain:
Suggestion is to merge this (or to remove `openjdk-11-jdk`) as Solr is started separately by `docker-compose.yml` through the [seek-solr](https://github.com/FAIRdom/solr-seek-docker/blob/master/Dockerfile) image which (if rebuilt) indirectly now will start [solr 8.9 using openjdk 11](https://github.com/docker-solr/docker-solr/blob/a1a59aa9d3ef286fe1e046ddfd168c3c34720660/8.9/Dockerfile). See #648

stuzart:
I'll merge this, and then also update to use `openjdk-11`. It will be ok as long as the container is only used with `docker-compose`.
Once synchronised with the `master ` branch, it will include updated configurations and `docker-compose` files to use an upgraded version of `seek-solr`

PR #659, opened 2021-07-01
Publications upload via versions

abecam:
Changes for the publications as discussed, without changing the Seek library now.
Also several fixes for the tests that failed here.


app/serializers/publication_serializer.rb
```diff
@@ -28,6 +28,36 @@ class PublicationSerializer < PCSSerializer
...
+        blobs = [requested_version.content_blob].compact
+      else
+        blobs = []
+      end
+
+      blobs.map { |cb| convert_content_blob_to_json(cb) }
+    end
+  end
+
+  def convert_content_blob_to_json(cb)
```
stuzart, line 47: Perhaps this method could be moved from ContributedResourceSerializer to higher up the hierarchy to avoid duplication

test/functional/publications_controller_test.rb
```diff
@@ -339,7 +343,10 @@ def test_title
   end
 
   test 'should show the publication with unspecified publication type as Not specified' do
-    get :show, params: { id: publications(:no_publication_type) }
+    publication = Factory(:publication, title: 'Publication without type')
+    publication.publication_type = nil
+    publication.save(validate: false)
```
stuzart, line 348: you could just put this in the factory call `Factory(:publication, title: 'Publication without type', publication_type: nil)`

test/fixtures/publications.yml
```diff
@@ -1,60 +1,61 @@
-# Read about fixtures at http://ar.rubyonrails.org/classes/Fixtures.html
```
stuzart, line 1: if these have now all be replaced with a Factory, may was well just remove all these rather than leaving commented out.

test/fixtures/relationships.yml
```diff
@@ -1,16 +1,16 @@
 # Relationships
```
stuzart, line 1: once again, may as well just remove this file if they are all commented out and no longer needed

test/functional/publications_controller_test.rb
abecam: I actually had to revert that after running the tests. Setting publication_type in the factory creation does not pass the validation. It is possible to define a trait in the factory to remove the validation, but as it is used only here, I don't think it would make sense.

abecam:
Aside the Factory(:publication, title: 'Publication without type', publication_type: nil) that I reverted as explained (does not pass validation), I applied all the changes.

PR #687, opened 2021-08-18
Single page basic

vdkkia:


app/controllers/single_pages_controller.rb
```diff
@@ -0,0 +1,57 @@
+class SinglePagesController < ApplicationController
+  include Seek::AssetsCommon
+  before_action :login_required, except: [:render_item_detail]
```
fbacall, line 3: Maybe `project_membership_required` is more appropriate?

app/controllers/single_pages_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @assay = Assay.new({})
+
+    respond_to do |format|
+      format.html
+    end
+  end
+
+  def render_item_detail
+    begin
+      raise Exception, "not logged in" if !User.logged_in?
```
fbacall, line 22: This should be checked by a `before_action`

app/controllers/single_pages_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def single_page_enabled
+    unless Seek::Config.project_single_page_enabled
+      flash[:error]="Not available"
+      redirect_to Project.find(params[:id])
+    end
+  end
+
+  def project_folders
+    project_folders = ProjectFolder.root_folders(@project)
```
fbacall, line 49: Would be nicer if this worked like: `@project.folders.roots`

app/views/assays/_show.html.erb
```diff
@@ -0,0 +1,121 @@
+<% single_page ||= false %>
```
fbacall, line 1: I think this is not a good way of doing things, it involves adding this option to too many partials.

Instead, you could set a variable in the controller, and use the `helper_method` function to make it globally available to views, like what is done for `controller_model`:
https://github.com/seek4science/seek/blob/6428fe4908958b1fe1e2a142010bf2e2cf3d820c/app/controllers/application_controller.rb#L128-L135

app/views/assays/show.html.erb
```diff
@@ -1,121 +1 @@
...
-  </div>
-
-  <div class="col-md-3 col-sm-4">
-    <%= render :partial=>"assets/resource_main_content_right",:locals => { :resource => @assay } %>
-  </div>
-</div>
-
-
-<%= render :partial=>"general/items_related_to",:object => @assay %>
+<%= render :partial=>"show" -%>
```
fbacall, line 1: I don't really like this way. I think it would be better if you created a special minimal layout (https://github.com/seek4science/seek/blob/f8b05ca07bbce8b8db2c4370d7ce288ee515c7de/app/views/layouts/application.html.erb) and just rendered the view in an iframe or something.

app/views/investigations/_form.html.erb
```diff
@@ -1,3 +1,6 @@
+<% object ||= nil %>
+<% showDetails ||= 1 %>
```
fbacall, line 2: Use `snake_case` for anything in Ruby that is not a class

app/views/assets/_tags_box.html.erb
```diff
@@ -1,5 +1,8 @@
 <%
-   entity=controller_name.singularize
+   # Make this partial work when calling from views of different controllers by passing 'ctrl_name' param
+   ctrl_name||=nil
+   resource_id||=nil
+   entity=ctrl_name&.singularize||controller_name.singularize
```
fbacall, line 5: I think all these things are a consequence of the decision to render all types of views from the same controller.

I think you would probably find things easier if you mixed in a single action into each asset type's controller.

We do this for previews:
https://github.com/seek4science/seek/blob/09f6b68b8a33bae4bacf733f94af94dc8ad728c3/lib/seek/preview_handling.rb

fbacall (commented):
I think you should revisit how the right-hand side content is loaded. It may make things simpler and require fewer changes to existing views and partials.

app/views/single_pages/_assay.html.erb
```diff
@@ -0,0 +1,15 @@
...
+                <a data-toggle="tab" href="#assay-related-items">
+                    <i class="glyphicon glyphicon-th-list"></i>&nbsp;Related items
+                </a>
+            </li>
+        </ul>
+        <div class="tab-content">
+            <div id="assay-details" class="tab-pane fade in active" style="padding:10px"></div>
+            <div id="assay-related-items" class="tab-pane fade" style="padding:10px"></div>
+        </div>
+    </div>
```
stuzart, line 15: these files are all very similar with only a couple of ids or hrefs based on the resource type. Could be put into a generic partial that derives the naming form the resource itself or is passed in as a local attribute. It would make it easier to maintain and remove the need to add a new page whenever a new resource type is added

app/views/investigations/_form.html.erb
```diff
@@ -12,18 +15,18 @@
 
 <%= render partial: 'custom_metadata/custom_metadata_type_selection', locals:{f:f, resource:@investigation} %>
 
-<%= render :partial => 'projects/project_selector', :locals => {:resource => @investigation,:allow_nil=>false} if show_form_manage_specific_attributes? -%>
+<%= render :partial => 'projects/project_selector', :locals => {:resource => @investigation,:allow_nil=>false} if show_form_manage_specific_attributes? && showDetails==1 -%>
 
 <%= render partial: 'custom_metadata/custom_metadata_attribute_input', locals:{f:f,resource:@investigation} %>
 
-<%= render :partial => "scales/scales_selector",:locals=>{:object=>@investigation} if show_scales? %>
+<%= render :partial => "scales/scales_selector",:locals=>{:object=>@investigation} if show_scales? && showDetails==1  %>
```
stuzart, line 22: why is showDetails flag using 1 or 0 rather than just true or false ?

app/views/assays/show.html.erb
stuzart, line 1: this seems to be added to allow the full show page for an item to be presented within the folder view when clicking on an item. I'm not sure this is what we want, and could lead to other problems such as behaviour when clicking the buttons, or issues with the back button.
You could instead a simpler preview of the item, with a link to go to the full show page from there (perhaps in a modal box)

stuzart (changes requested):
There seems to be a lot of changes that cuts across many pages, controlled by various flags overloading the existing pages. It appears to be a side affect of going through the single page controller
I'm worried that this may introduce future problems, and would be worth investigating a cleaner solution that doesn't have so much cross-cutting impact and better separation of concerns

vdkkia (commented):
The single page is checked with a helper method now

app/controllers/application_controller.rb
```diff
@@ -637,4 +638,17 @@ def determine_custom_metadata_keys
     end
     keys
   end
+
+  def displaying_single_page?
+    instance_variable_get("@single_page")||false
```
stuzart, line 643: my comment on the call, related to this. Why not just `@single_page || false` - I don't think instance_variable_get is needed

config/routes.rb
```diff
@@ -701,6 +701,14 @@
     end
   end
 
+   ### SINGLE PAGE
+
+  resources :single_pages do
+    member do
+      get "/render_item_detail/:id/type/:type" => "single_pages#render_item_detail"
```
fbacall, line 708: I still think it would be easier if you did this through each type's controller instead of a single controller.

You would not have to check types and could use the existing methods to find and authorize assets.

vdkkia: Doing this through each type's controller works but it's not possible to set `@single_page` in order to use `displaying_single_page` helper method in the views. Is there any solution for this?

fbacall: Why not? Can't you just set it in the same way?

vdkkia: I can set it as it is. I tried to implement your approach but, I noticed that the type controller gets active when for example send this Ajax request: `/investigations/1` (then it shows `@single_page` as undefined)

fbacall: You would need to call `/investigations/1/render_item_detail` and have a `before_action ... only: :render_item_detail`.

We have a similar feature for "previews", which is available for all asset + ISA types which you might want to look at*

https://github.com/seek4science/seek/blob/master/lib/seek/preview_handling.rb
https://github.com/seek4science/seek/blob/master/lib/seek/assets_common.rb#L7
https://github.com/seek4science/seek/blob/master/config/routes.rb#L94
https://github.com/seek4science/seek/blob/master/config/routes.rb#L108

*Although the behaviour is slightly wrong/weird since the route ends up like `/data_files/preview?id=123`, and it does the authorization in the view.

PR #705, opened 2021-09-22
Seek 1.11 licenses

stuzart:


config/application.rb
```diff
@@ -76,5 +76,8 @@ class Application < Rails::Application
 
     # Ignore translation overrides when testing
     config.i18n.load_path += Dir[Rails.root.join('config', 'locales', 'overrides', '**', '*.{rb,yml}')] unless Rails.env.test?
+
+    config.recommended_data_licenses = []
```
stuzart, line 80: this I think will overwrite the config whenever the app is restarted. The defaults can instead go in the default settings @ /config/initializers/seek_configuration.rb - and should the default be the fall-back setting ?

app/views/workflows/edit.html.erb
```diff
@@ -26,7 +26,7 @@
       <p class="help-block">This field is used to indicate to users what level of stability they can expect from the workflow.</p>
     </div>
 
-    <%= render partial: 'assets/license_selector', locals: { resource: @workflow, license_group: :software, default_license: 'Apache-2.0' } %>
+    <%= render partial: 'assets/license_selector', locals: { resource: @workflow, license_group: :software, recommended_licenses: Seek::Config.recommended_software_licenses, default_license: 'Apache-2.0' } %>
```
stuzart, line 29: the recommended licenses can be determined inside the partial, based on the license_group. currently this only affects workflows

lib/seek/config.rb
```diff
@@ -15,6 +15,14 @@ def dm_project_name_fallback
       project_name
     end
 
+    def recommended_data_licenses_fallback
+      ['CC-BY-4.0']
+    end
+    
+    def recommended_software_licenses_fallback
+      ['CC0-1.0']
```
stuzart, line 23: workflow uses Apache-2 as the default license

app/views/workflows/edit.html.erb
alaninmcr: I agree with Stuart

stuzart:
merged

PR #724, opened 2021-10-01
Unregistered author improvements

### LINKED ISSUE #681: Improve "other creators"/non-registered author credit
Currently it is just a text field. In order to support proper semantic markup, automatic generation of citations etc. this should be a more structured format, allowing name, affiliation, ORCID to be specified per creator without them requiring a SEEK profile.

Related: #391

fbacall:
To fix #681


app/assets/stylesheets/new_styles.scss
```diff
@@ -1073,3 +1073,37 @@ span.disabled {
 .citation-box {
   overflow-x: auto;
 }
+
+a.orcid-link {
+  background-image: url(image-path('logos/orcid_16x16.png'));
```
stuzart, line 1078: would be good if this image path was taken from the image dictionary. But I see there other hard-coded cases in this file so can update after merging

app/models/assets_creator.rb
```diff
@@ -1,9 +1,73 @@
...
+
   update_rdf_on_change :asset
 
-  default_scope { order(:id) }
+  default_scope { order(:pos) }
+
+  def family_name
+    creator ? creator.last_name : super
+  end
+  alias_method :last_name, :family_name
```
stuzart, line 17: looks like we need some refactoring for consistent use of family/given name over first and last. Maybe not for this PR but worth adding an issue

PR #728, opened 2021-10-06
Datahub template

vdkkia:
There's the ability to show/create templates in this branch. It's supposed to be used in the full version of the **single page** (Sample types will be created based on the templates).
Since there are not many things that users can do with the templates in the Seek base, an on/off switch is considered for **_Templates_** in the server admin panel.


app/controllers/templates_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+          format.html { render action: 'edit', status: :unprocessable_entity }
+          format.json { render json: @template.errors, status: :unprocessable_entity}
+        end
+      end
+    end
+  
+    def destroy
+      respond_to do |format|
+      if @template.can_delete? && @template.destroy
+        format.html { redirect_to @template,location: templates_path, notice: 'Template was successfully deleted.' }
```
stuzart, line 64: hardcoded "Template", but I see later on there is a I18n definition that could be used. Same a couple of lines below

app/views/templates/edit.html.erb
```diff
@@ -0,0 +1,3 @@
+<h1>Editing Template</h1>
```
stuzart, line 1: hard coded "Template" rather than I18n

app/views/templates/new.html.erb
```diff
@@ -0,0 +1,39 @@
+<% tab = @tab || 'manual' %>
+
+<h1>New Template</h1>
+
+<p>
+  Templates can be defined manually through the form below or based on repository templates.
```
stuzart, line 6: hard coded "Template" rather than I18n

stuzart (changes requested):
I found a few cases where the t('template') locale wasn't being used, and hard-coded text instead. You might want to check if there any other cases I missed.

There is potential to re-use and combine with the sample-types more, to avoid as much duplication, but this I think it is a later refactoring task once things have settled. 

This is quite a big PR so I will have another scan over later and may add more comments.

app/controllers/templates_controller.rb
vdkkia: This issue and the other similar issues have been solved via the last commit.

PR #767, opened 2021-10-29
I18n asset gatekeeper

whomingbird:


stuzart (changes requested):
all just a/an problems. Can maybe just be replaced with 'the' (better might be "one of the" if plural but I think it is overkill).

PR #921, opened 2022-03-18
Remove vln

stuzart:
Removal of old virtual liver related code, which were otherwise disabled


app/models/admin_defined_role_project.rb
```diff
@@ -20,9 +20,10 @@ def queue_update_auth_table
     AuthLookupUpdateQueue.enqueue(person)
   end
 
-  def project_belongs_to_person_or_subprojects
-    unless (Seek::Config.project_hierarchy_enabled && (person.try(:projects_and_descendants) || []).include?(project)) || (person.try(:projects) || []).include?(project)
-      errors.add(:project,"the project must be one the person is a member of")
+  def person_must_be_in_project
+    unless person && project && person.projects.include?(project)
```
fbacall, line 24: Could use `project.has_member?(person)` here

app/models/person.rb
```diff
@@ -136,7 +136,12 @@ class Person < ApplicationRecord
   def person
     self
   end
-    
+
+  def projects
+    # you can't access these through .projects nested associations until they have been saved
+    work_groups.collect(&:project).uniq | group_memberships.collect { |gm| gm.work_group.project }
+  end
+
```
fbacall, line 144: What is the reason behind this change?

fbacall (commented):
Left some comments, otherwise looks fine

app/models/person.rb
stuzart: it's not actually a change, but was there before coming from the project hierarchy stuff, so just got moved over

PR #926, opened 2022-03-22
error on batch permission change page when the user has a programme #925

whomingbird:


app/views/assets/sharing/batch_sharing_permission_preview.html.erb
```diff
@@ -6,7 +6,7 @@
   all_investigations_hash[:children] = []
 
   #remove the public items which don't need change sharing permissions
-  all_related_items_hash = get_related_resources(current_user.person).except!('Project', 'Institution','Publication')
+  all_related_items_hash = get_related_resources(current_user.person).except!('Project', 'Institution','Publication','Programme')
```
fbacall, line 9: I would find a solution that uses `Seek::Util.authorized_types`. Otherwise, it would break again if we add another type without authorization in the future.

PR #929, opened 2022-03-23
More vln removal

stuzart:
Finals parts of removing VLN behaviour

- removed the need to override Person.can_view?
- removed the need to override Person.projects for the tests & factories to work
- removed assay->institution
- fixed MessageLog details serialization, to handle legacy cases and avoid future problems


lib/seek/roles/project_related_roles.rb
```diff
@@ -157,7 +157,7 @@ def is_in_any_gatekept_projects?
         # any for projects that have been removed, and resolving the mask
         def resolve_admin_defined_role_projects
           admin_defined_role_projects.each do |role|
-            role.destroy unless projects.include?(role.project)
+            role.destroy unless group_memberships.collect(&:project).include?(role.project)
```
fbacall, line 160: Is it done this way because of the issue where the association doesn't work on unsaved associations?

app/models/programme.rb
```diff
@@ -18,7 +18,7 @@ class Programme < ApplicationRecord
   has_many :projects, dependent: :nullify
   has_many :work_groups, through: :projects
   has_many :group_memberships, through: :work_groups
-  has_many :people, -> { order('last_name ASC').distinct }, through: :group_memberships
+  has_many :people, through: :group_memberships
```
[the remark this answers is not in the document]
stuzart: oops, yes I'll fix that

lib/seek/roles/project_related_roles.rb
stuzart: yes, otherwise you can't add somebody to a project and assign a role without saving in between. At first I updated the test to do an intermediate save, but changed my mind and decided this was safer ( https://github.com/seek4science/seek/pull/929/commits/a8a92d22d1d7101aebffe7956781cf23e7ab0546 )

Issue #927, opened 2022-03-22
update publication types 


whomingbird:
- [ ] rewording 

Masters -> Master's
Bachelors -> Bachelor's
PhD -> Doctoral

- [ ]  add "Diplom Thesis” as a new publication type;

- [ ] add a rake task to update publication types of the running SEEK.

stuzart:
@whomingbird could we do this on the seek-1.12 branch ?

PR #933, opened 2022-03-24
update publication types #927

whomingbird:


lib/tasks/seek_dev.rake
```diff
@@ -291,9 +291,21 @@ namespace :seek_dev do
     output.close
   end
 
+  task update_thesis_related_publication_types: :environment do
```
stuzart, line 294: this needs adding to the upgrade task, and should only print out those that needed updating (i.e haven't already been updated)

stuzart (changes requested):

lib/tasks/seek_upgrades.rake
```diff
@@ -131,4 +132,31 @@ namespace :seek do
       puts "Database adapter is: #{ActiveRecord::Base.connection.instance_values["config"][:adapter]}, doing nothing"
     end
   end
+
+
+  task(update_thesis_related_publication_types: [:environment]) do
+    puts 'Updating publication types ...'
+
+    unless PublicationType.find_by(title:"Masters Thesis").nil?
+      PublicationType.find_by(key:"mastersthesis").update_attributes(title:"Master's Thesis")
```
stuzart, line 141: not a blocker, but to make you aware. With Rails 6 (on the master branch), `update_attributes `is deprecated and should instead use just `update`. It doesn't affect 1.12 though, so I can update this when we merge.

Issue #941, opened 2022-03-30
`Seek::Config.site_base_host` should be stripped of sub-URIs


fbacall:
In the admin section it says that:
> It should only be set to the host, but not include any additional paths if running under a sub uri. This is required cases where the host cannot be determined from the request.

We should ensure this is the case by stripping any paths before the setting is saved, otherwise it could lead to wrong URLs being generated (`http://example.com/seek/seek/data_files`)

Issue #940, opened 2022-03-30
Sub-URI not included in manually crafted URLs using `site_base_host`


fbacall:
Lots of places manually generate URIs by using `Seek::Config.site_base_host` and concatenating some strings. This will not include any sub-URIs.

This is done in lots of places throughout the code for generating RDF URIs etc.

Need to go through and check all usages of `Seek::Config.site_base_host`.

fbacall:
Fixed by #946

Issue #938, opened 2022-03-28
Sub-URI issues with hardcoded URLs in JS


fbacall:
Lots of places have root-relative paths hard-coded in the JS like this:
https://github.com/seek4science/seek/blob/seek-1.12/app/assets/javascripts/project_folders.js#L108
https://github.com/seek4science/seek/blob/seek-1.12/app/assets/javascripts/openbis.js#L59 (Also this route is non-existent...)

this won't work if the app is deployed under a sub-URI. Need to replace these with something like this:
https://github.com/seek4science/seek/blob/seek-1.12/app/assets/javascripts/sharing.js.erb#L79

If there are URL parameters required (IDs etc.) these can be added using placeholders that are dynamically substituted at run-time in the JS, e.g.
```
var showDatasetFilesPath = '<%= Rails.application.routes.url_helpers.show_dataset_files_openbis_endpoint_openbis_dataset_path('-endpoint_id-', '-dataset_id-') %>';

...

var path = showDatasetFilesPath.replace('-endpoint_id-', endpointId).replace('-dataset_id-', permId);
```

stuzart:
can you `include UrlHelper` that makes it a bit cleaner and just call `show_dataset_files_openbis_endpoint_openbis_dataset_path` in the example above?

fbacall:
I've tried to avoid including routes as much as possible, because I've had obscure errors where it clobbers some Rails methods and causes strange behaviour. It's probably OK to include in JavaScript though, and is done in other files.

fbacall:
Fixed by #946

PR #946, opened 2022-03-31
Sub-URI support

### LINKED ISSUE #934: Under docker, avatars don't seem to load
Seems to be something to do with Rails serving static assets in `production` mode. The env var `RAILS_SERVE_STATIC_FILES` controls this behaviour. It may have been that this was defaulting to "true" prior to the Rails 6 changes.

This also indicates that there is something potentially wrong with the nginx config, since the request should not be getting to the app server in the first place - nginx should be serving everything under `/public`

fbacall:
Docker changes to allow nginx to serve static files when running with a sub-URI.

Lots of changes to how SEEK generates URLs to ensure the sub-URI is respected.

Fixes #934, #938, #940, #941


app/controllers/culture_growth_types_controller.rb
```diff
@@ -0,0 +1,16 @@
+class CultureGrowthTypesController < ApplicationController
```
stuzart, line 1: I don't see where this is need or used, other than the routes ??

stuzart: is it to allow polymorphic_url to work for all cases in the rdf generation ?

vendor/assets/stylesheets/jstree.scss
```diff
@@ -403,7 +403,7 @@
 }
 .jstree-default .jstree-node,
 .jstree-default .jstree-icon {
-  background-image: url("32px.png");
+  background-image: url(image-path("32px.png"));
```
stuzart, line 406: since this is editing a `vendor/assets` can it be moved to `app/assets` ?

app/controllers/culture_growth_types_controller.rb
fbacall: Yes exactly

vendor/assets/stylesheets/jstree.scss
fbacall: Hmm, maybe. I didn't really like the solution of editing the CSS, but couldn't find any other way around it.

I got the idea from https://makandracards.com/makandra/29567-managing-vendor-libraries-with-the-rails-asset-pipeline#section-fixing-font-paths

I did wonder if there's a way to automatically do this as part of the asset-pipeline compilation, or how it works for the old rails-assets gems

stuzart: actually, there are plenty of other cases where files here have been edited

Issue #886, opened 2022-02-15
Remove Scales


stuzart:
an old feature that is no longer used or maintained.

PR #947, opened 2022-03-31
Remove scales #886

fbacall:


app/models/scale.rb
```diff
@@ -1,53 +0,0 @@
...
-
-  # attr_accessible :image_name, :key, :pos, :title
-  alias_attribute :name, :title
-
-  default_scope -> { order("pos ASC") }
-
-  validates_presence_of :title,:key,:image_name
-  validates_uniqueness_of :title,:key,:image_name
-
-  after_destroy :remove_annotations
```
stuzart, line 12: could there be annotations left hanging around after the scales related tables are dropped, and does it matter ?

stuzart (commented):
Looks good, other than the comment I added that needs double checking

app/models/scale.rb
fbacall: Yeah you're right, I'll add a task to clear them out

stuzart (commented):
Looks good, other than the comment about old annotations hanging around needs double checking

PR #951, opened 2022-04-04
Master 876

alaninmcr:
Removed lib and api_helper based upon Finn's re-opening


app/serializers/avatar_obj_serializer.rb
```diff
@@ -1,5 +1,7 @@
 class AvatarObjSerializer < BaseSerializer
   attribute :avatar do
-    avatar_href_link(object)
+    uri = nil
+    uri = "/#{object.class.name.pluralize.underscore}/#{object.id}/avatars/#{object.avatar.id}" unless object.avatar.nil?
```
fbacall, line 4: This will need fixing when #946 is merged

stuzart: should be able to replace with `polymorphic_path(object,avatar)`

app/serializers/base_serializer.rb
```diff
@@ -123,6 +122,18 @@ def show_policy?
     return respond_to_policy && respond_to_manage && can_manage
   end
 
+  def determine_submitter(object)
+    # FIXME: needs to be the creators for assets
```
stuzart, line 126: this comment is redundant and can be removed

app/serializers/base_serializer.rb
```diff
@@ -123,6 +122,18 @@ def show_policy?
...
+    return object.owner if object.respond_to?('owner')
```
stuzart, line 127: I'm not sure this is needed, everything should have a contributor rather than owner, of if not have an alias or be updated

app/serializers/base_serializer.rb
```diff
@@ -123,6 +122,18 @@ def show_policy?
...
+    result = object.contributor if object.respond_to?('contributor') && !object.is_a?(Permission)
+    if result
+      return result if result.instance_of?(Person)
+      return result.person if result.instance_of?(User)
```
stuzart, line 131: also no longer needed, you can just return the result.  It will always be a Person here

stuzart (changes requested):

app/serializers/base_serializer.rb
alaninmcr: Comment removed

app/serializers/avatar_obj_serializer.rb
alaninmcr: Have not changed. Will wait for 946

app/serializers/base_serializer.rb
alaninmcr: This is a more general serialization issue and IMO should be raised as one

app/serializers/avatar_obj_serializer.rb
fbacall: this can now be changed to
`polymorphic_path([object, object.avatar]) unless object.avatar.nil?`

PR #988, opened 2022-04-28
API testing

### LINKED ISSUE #884: Improve and simplify how API tests work
It's a big overhead to extend or add new API tests, and difficult to fix and track down ones that fail. They don't always test against the min and max definitions. They are complex to understand how they work.

fbacall:
Refactor how API-related tests work.

Both read and write API endpoints are tested together under `test/integration/api` rather than being split across functiona/integration tests.

request/response templates are in `test/fixtures/json`.

Request templates are used to test create and update methods and form the request body. They are evaluated using instance variables declared in the test `setup` method.

Response templates are used to check that the read responses are structured in an expected way. They are evaluated using an object `res` (not settled on this name). All parts of the response are checked, so the templates are structured to cope with all the variable fields.

Fixes #884

### DIFF
--- a/test/functional/data_files_controller_test.rb
+++ b/test/functional/data_files_controller_test.rb
@@ -17,14 +17,6 @@ class DataFilesControllerTest < ActionController::TestCase
   def setup
     login_as(:datafile_owner)
   end
-  #
-  # def rest_api_test_object
-  #   # by TZ for some reason depending on tets order user was no longer logged
-  #   login_as(:datafile_owner) unless User.current_user
-  #   @object = data_files(:picture)
-  #   @object.annotate_with 'tag1'
-  #   @object
-  # end
 
   def test_title
     get :index
--- a/test/functional/presentations_controller_test.rb
+++ b/test/functional/presentations_controller_test.rb
@@ -10,12 +10,6 @@ class PresentationsControllerTest < ActionController::TestCase
     login_as Factory(:user)
     @project = User.current_user.person.projects.first
   end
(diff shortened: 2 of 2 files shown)


test/functional/data_files_controller_test.rb
```diff
@@ -18,14 +17,14 @@ class DataFilesControllerTest < ActionController::TestCase
...
-    @object = data_files(:picture)
-    @object.annotate_with 'tag1'
-    @object
-  end
+  #
+  # def rest_api_test_object
+  #   # by TZ for some reason depending on tets order user was no longer logged
+  #   login_as(:datafile_owner) unless User.current_user
+  #   @object = data_files(:picture)
+  #   @object.annotate_with 'tag1'
```
stuzart, line 25: just delete this, rather than comment out ?

test/functional/presentations_controller_test.rb
```diff
@@ -4,19 +4,18 @@
...
-    @object = Factory :presentation, contributor: User.current_user.person
-    @object.annotate_with 'tag1'
-    @object
-  end
+  #
+  # def rest_api_test_object
+  #   @object = Factory :presentation, contributor: User.current_user.person
+  #   @object.annotate_with 'tag1'
+  #   @object
+  # end
```
stuzart, line 18: delete ?

stuzart (approved):
just a couple of commented blocks that need removing. 
I didn't check every large diff so you may want to check if there are any more

Issue #992, opened 2022-04-29
Ordering option is wrongly available


alaninmcr:
The option to order items is shown even though the user does not have full permission.

alaninmcr:
For example I supposedly cannot even see the investigations in https://fairdomhub.org/projects/106 but can order them :-O

alaninmcr:
Similarly https://fairdomhub.org/projects/250 allows me to manage the project when I should not have the rights

stuzart:
you're an administrator

alaninmcr:
Found the underlying issue. Need to check that user can manage the resources being ordered.

Added code to do this to fairdomhub-992

Probably not the correct branch :)

stuzart:
Does it make sense that the items being ordered need manage rights? I'd say it should be that just edit rights on the container that they are being ordered within is required?

Make sure you fix it off the seek-1.12 branch, no the fairdomhub branch.

alaninmcr:
Bit confused now - though code is fairly easy to change.

Container -> edit?
Contained -> manage?

or do you want

Container -> edit
Contained -> any rights (possibly at least view)

stuzart:
I'd say the 2nd case. This matches the rights for adding things to an Assay

alaninmcr:
Changed to make edit (for container) and edit for contained.

PR #994, opened 2022-05-04
Fixes to ordering

alaninmcr:
Should deal with #992


stuzart (commented):
They change means you can only order if you can edit both the container and the sub items, which is probably too restrictive. By ordering things you are only really changing the container, not the items in it

PR #1000, opened 2022-05-10
Isa json export

vdkkia:
**ISA JSON export** is developed on top of Single pages advanced


config/image_files/image_file_dictionary.yml
```diff
@@ -129,6 +129,7 @@ saved_search: 'crystal_project/32x32/actions/find.png'
 saved_search_avatar: 'crystal_project/32x32/actions/find.png'
 search: 'famfamfam_silk/eye.png'
 seek_logo: 'fairdom-seek-logo.svg'
+elixir_belgium_logo: 'elixir.png'
```
stuzart, line 132: would be better if the png was renamed elixir-beligium.png

lib/ebi/ols_client.rb
```diff
@@ -66,6 +66,20 @@ def self.ontology_keys
       @ontology_keys ||= ontologies.map { |ontology| ontology.dig('config', 'namespace') }.sort.compact
     end
 
+    def fetch_ontology_reference(ontology_id)
+      Rails.cache.fetch("ebi_ontology_terms_#{ontology_id}") do
```
stuzart, line 70: key should say reference rather than terms

lib/seek/samples/attribute_type_handlers/cv_attribute_type_handler.rb
```diff
@@ -5,7 +5,7 @@ class CVAttributeTypeHandler < BaseAttributeHandler
         class MissingControlledVocabularyException < AttributeHandlerException; end
 
         def test_value(value)
-          fail "'#{value}' is not included in the controlled vocabulary" unless controlled_vocab.includes_term?(value)
+          fail "'#{value}' is not included in the controlled vocabulary" unless controlled_vocab.includes_term?(value) || controlled_vocab.source_ontology.present?
```
stuzart, line 8: this looks wrong. Even if the CV is from an ontology, the value should still be present in the list of terms to be valid

test/unit/samples/cv_attribute_type_handler_test.rb
```diff
@@ -24,4 +24,12 @@ class CVAttributeTypeHandlerTest < ActiveSupport::TestCase
       assert handler.validate_value?('Granny Smith')
     end
   end
+
+  test 'bypass validation for ontology terms' do
```
stuzart, line 28: shouldn't be bypassing validation for ontology terms

stuzart (changes requested):
first pass, will go over again later as there is a lot to review.

check for other hard-coded paths as I've probably missed several

lib/seek/samples/attribute_type_handlers/cv_attribute_type_handler.rb
vdkkia: What do you suggest if we want to accept inputs that are not in the list of **terms**?
**Requirement**: The users should be able to either choose from terms or input a **custom** value.

stuzart: if it is not in the list of terms, then it cannot be resolved to it's associated URI, which will then lead to other problems. It also prevents it behaving as a controlled vocabulary.

If it is an essential requirement then it would be better to have a flag on the SampleControlledVocab, which can be used to override the normal behaviour where the user wants to allow arbitary terms.

app/controllers/single_pages_controller.rb
```diff
@@ -35,6 +33,52 @@ def project_folders
     project_folders
   end
 
+  def ontology
+    begin
+      labels = (SampleControlledVocab.find(params[:sample_controlled_vocab_id])
+      &.sample_controlled_vocab_terms || [])
+      .where("LOWER(label) like :query", query: "%#{params[:query].downcase}%")
+      .select("label").limit(params[:limit] || 100)
```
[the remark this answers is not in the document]
vdkkia: Fixed

app/views/isa_studies/_source_material.html.erb
```diff
@@ -0,0 +1,79 @@
...
+        $j(document).ready(function () {
+            $j('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+                $j.fn.dataTable.tables( {visible: true, api: true} ).columns.adjust();
+            });
+            let dt = <%= sanitize(dt_data(sample_type).to_json) %>
+            sourceDynamicTable = new $j.dynamicTable('#source-material-table')
+            const elem = $j("#btn_save_source")
+            const options = {
+               ajax:{
+                    url: `/single_pages/${pid}/dynamic_table_data/`,
```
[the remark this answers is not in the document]
vdkkia: Is it good to convert it to 
`url: '<%= single_pages_path %>' + '/${pid}/dynamic_table_data/',
`
?

PR #1013, opened 2022-05-20
Seek 1.12 gatekeeper improvement

whomingbird:
- after the gatekeeper makes a decision (approve or reject), the publishing state of the resource for the contributor is not updated. For the contributor of the resource `is_waiting_approval?= true`, it should be updated to PUBLISHED or REJECTED

-  in 'requested_approval_assets' page, the recently updated items should be listed at the top.

-  should be able to open the resource link in a new tab on the 'requested_approval_assets' page ( better UX)

-  better UI of requested_approval_assets page. ( https://hostname/people/id/requested_approval_assets)
- 
-  bug fix: After `reject/decide late` decision is made, the approval is not working;  the error message is "Items that were decided to make published but were not published due to an unforeseen problem" 

-  The resource can be re-published after it is rejected by the gatekeeper before.  The "publish" button will appear again in the "Action" menu when the resource is updated.

### DIFF
--- a/lib/seek/permissions/publishing_permissions.rb
+++ b/lib/seek/permissions/publishing_permissions.rb
@@ -42,10 +42,13 @@ module Seek
             policy.access_type = Policy::ACCESSIBLE
             policy.sharing_scope = Policy::EVERYONE # anything but ALL_USERS
             policy.save
-            # FIXME: may need to add comment
+
             resource_publish_logs.create(publish_state: ResourcePublishLog::PUBLISHED,
                                          user: User.current_user,
                                          comment: comment)
+
+            resource_publish_logs.find_or_initialize_by(publish_state: ResourcePublishLog::WAITING_FOR_APPROVAL, resource_id: self.id, user_id: self.contributor.id)
+                                 .update(publish_state:ResourcePublishLog::PUBLISHED, comment:comment)
             touch
           end
         else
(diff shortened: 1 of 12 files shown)


lib/seek/permissions/publishing_permissions.rb
```diff
@@ -42,10 +51,15 @@ def publish!(comment = nil, force = false)
...
             policy.sharing_scope = Policy::EVERYONE # anything but ALL_USERS
             policy.save
-            # FIXME: may need to add comment
+
             resource_publish_logs.create(publish_state: ResourcePublishLog::PUBLISHED,
                                          user: User.current_user,
                                          comment: comment)
+
+            unless contributor.is_asset_gatekeeper_of?(self)
+              resource_publish_logs.find_or_initialize_by(publish_state: ResourcePublishLog::WAITING_FOR_APPROVAL, user: contributor.user )
```
stuzart, line 60: I'm confused by this, because a log has already been created above. Couldn't this create a 2nd duplicate log ?

stuzart (changes requested):
there are issues to do with previous logs be modified or deleted, and logs being made associated to the wrong user making the action. The logging is there to keep an audit incase we are asked about things being unexpectedly published

To get the current state, you should just be able to check the last log made.

stuzart:
replaced by #1027

PR #1023, opened 2022-06-01
Alternative installation guide using separate `seek` user, RVM alias,…

fbacall:
… and system passenger.

### DIFF
--- a/tech/install-production.md
+++ b/tech/install-production.md
@@ -24,25 +24,19 @@ To save time later there are also some additional packages to install:
 
     sudo apt-get install libapr1-dev libaprutil1-dev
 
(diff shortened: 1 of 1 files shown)


stuzart (approved):
Looks good, but I haven't tested it and assume it has been tested from a clean state.

I'm curious about permissions. Won't the www-data user need write permissions for logs, filestore, tmp files etc - assuming passenger runs as www-data spawned from the Apache process ?

fbacall:
I hadn't thought about that until you mentioned it, but it seems to "know" about the app user from the file permissions on the "startup file" (config.ru):
https://www.phusionpassenger.com/library/deploy/nginx/user_sandboxing.html#how-it-works

PR #1034, opened 2022-06-13
Single page improvements

vdkkia:


app/helpers/sample_types_helper.rb
```diff
@@ -91,7 +91,13 @@ def attribute_type_link(sample_type_attribute)
     end
 
     if sample_type_attribute.controlled_vocab?
-      type += ' - ' + link_to(sample_type_attribute.sample_controlled_vocab.title, sample_type_attribute.sample_controlled_vocab)
+			title = 
+			if sample_type_attribute.sample_attribute_type.ontology?
+				sample_type_attribute.sample_controlled_vocab.ols_root_term_uri
```
stuzart, line 96: I don't think we should be overriding the attribute title with the root URI, as it isn't very user friendly and most users won't recognise it. If anybody want a CV title to be the root term, they can set it as that

app/views/investigations/_new.html.erb
```diff
@@ -1,5 +0,0 @@
-<h1>New <%= t('investigation') %></h1>
-
-<%= form_for @investigation do |f| %>
-    <%= render :partial => "investigations/form", :locals => { :f => f, :action=>:new , :object => @investigation } -%>
-<% end -%>
```
stuzart, line 5: I'm unclear what this has been replaced with, and doesn't it just leave the file blank?

app/views/studies/_new.html.erb
```diff
@@ -1,15 +0,0 @@
...
-  <%= button_link_to("New #{t('investigation')}", 'arrow_right', new_investigation_path) -%>
-<% else %>
-  <h1>New <%= t('study') %></h1>
-
-  <div class="asset_form">
-    <%= form_for @study do |f| %>
-      <%= render :partial => "studies/form", :locals => { :f => f, :action=>:new} -%>
-    <% end -%>
-  </div>
-<% end -%>
```
stuzart, line 15: I'm unclear what this has been replaced with, and doesn't it just leave the file blank?
It will lose the prompt to provide an Investigation if none are already available.

stuzart (changes requested):

app/views/investigations/_new.html.erb
vdkkia: I thought these **_new** partials are from when the **study**/**investigation** create modals were shown on the single page. Moreover, I expected these files to be deleted...

stuzart: OK, you're right, I didn't notice they are partials and they had been deleted.

app/helpers/sample_types_helper.rb
floradanna: `I don't think we should be overriding the attribute title with the root URI` Just to clarify, we don't want to override sample_type attribute title. The PR is about CV title.

The idea is that IF a CV list comes from an ontology, the CV list title is the Ontology Root Term by default. In this way, the CV list made from Ontology become more reusable, since there is no custom naming involved. This would only apply in case of Ontology, not for a CV made from free text terms.

To make it more flexible, maybe we could allow "overriding" the default name (Ontology Root Term) if a user really wants to customise it.

@stuzart what do you think? should we discuss it with the community workers?

stuzart: The URI is intended to be machine readable and not for human consumption. From working with users in the past, we find they find being exposed to ontologies and semantic web disconcerting and puts them off. The controlled vocabulary has a title that should state what it is, but the URI is just cryptic. 

Further, the link it topsy turvey and confusing - it displays a URL but when you click the link it goes somewhere else - to the controlled vocab definition. That definition includes details about the CV and what ontology it came from.

The link is to a controlled vocabulary that has been populated from a partial ontology, not to the ontology itself.

Issue #1036, opened 2022-06-15
Table view column has misnamed Doi


alaninmcr:
The table views (from Browse) that allow a DOI, misspell it as Doi. That is incorrect and should be fixed.

Issue #1035, opened 2022-06-14
email mandatory but not as documented


alaninmcr:
It is mandatory to specify an e-mail when creating a person in the API. The API documentation (redoc) does not specify this.

PR #1047, opened 2022-06-21
Little bug fixes

stuzart:
Couple of small bug fixes -

- fixed schema for `personPost `to say email is required, also updated the `personResponse ` to include all required attributes #1035
- fixed rendered of **Doi** for index table columns, and used i18n to control displayed attributes generally #1036 ( also added assay technology type)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -150,6 +150,7 @@ en:
           other: "%{count} errors prohibited this %{model} from being saved"
         # The variable :count is also available
         body: "There were problems with the following fields:"
+        body: "There were problems with the following fields:"
     models:
       workflow_class: Workflow Type
     attributes:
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -150,6 +150,7 @@ en:
           other: "%{count} errors prohibited this %{model} from being saved"
         # The variable :count is also available
         body: "There were problems with the following fields:"
+        body: "There were problems with the following fields:"
```
fbacall, line 153: Duplicate?

config/locales/en.yml
```diff
@@ -180,6 +181,11 @@ en:
     modify_immutable_remote_error: 'You cannot make changes to an imported repository.'
     modify_immutable_error: 'You cannot make changes to a frozen version.'
 
+  index-table-columns:
+    assay_type_uri: 'Assay Type'
+    technology_type_uri: 'Technology Type'
+    doi: 'DOI'
+
```
fbacall, line 188: Is it worth using the Rails activemodel/record conventions to do this?

https://github.com/seek4science/seek/blob/054ef0f1646780c7e3ed47b55d3f15cb2b2b6753/config/locales/en.yml#L140-L143

https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

stuzart: I did wonder that after I'd changed it, but wasn't sure it would give any real advantage. But I wasn't aware it was already being used elsewhere so will update it to be consistent

config/locales/en.yml
stuzart: looks like it, but I'm not sure how, must have been a finger slip

Issue #862, opened 2022-02-03
Review the options available for the table view columns


stuzart:
Some items are missing (such as created at and update at), and there is some inconstencies between common attributes available for one type, and not another. Needs a review and a list created.

stuzart:
whilst working on #1036 I saw that the the way columns are inherited from superclasses doesn't quite work as expected, which is what is leading to inconsistencies and missing attributes. I will fix that, and also add any obvious missing attributes to resolve this issue

PR #1063, opened 2022-07-05
Improvements to table columns definitions for index pages

stuzart:
All relates to #862 and includes:

- Centrally controlled via a YAML file that can more easily be updated and reviewed
- A broad set of general defaults that are inherited from, to provide more consistency
- protection against accessing dis-allowed attributes
- moved some logic out of the views and into helpers
- Improved the formatting of some field types, such as times, countries and licenses
- removed some duplication from the condensed view, both now use the same code to render the value

I'm thinking of coming back to this in the near future to make the condensed view columns defined explicitly (it currently uses the first 3 default columns), which will make it a bit simpler and flexible


abecam (approved):
All ok.

Issue #1069, opened 2022-07-07
Share with programme option appears even if Programmes disabled


stuzart:
Button shows up on sharing form. Found as part of #573

Issue #573, opened 2021-04-19
Confusion over naming of Independent teams / projects


stuzart:
Sharing option list has independent teams (or project) to mean those without a Space (Programme), but that term is also used name a managed Space/Programme on fairdomhub and workflowhub

stuzart:
Standalone

PR #1070, opened 2022-07-07
Sharing form wording and programme button fix

stuzart:
Reworded Independent projects as Projects without a Programme ( #573 )
.. also fixed used of locale for People/Person, and capitalized for consistency with other views

Share with Programme button should only show if enabled #1069


abecam (approved):
All ok!

Issue #449, opened 2020-10-28
Changing sample type title attribute should expire RLI cache for samples


fbacall:
Or it would show the old title

PR #1083, opened 2022-07-12
Clear cached sample resource list item title when sample type changed

stuzart:
clears the cache in the sample type update job if refresh samples selected #449

### DIFF
--- a/app/jobs/sample_type_update_job.rb
+++ b/app/jobs/sample_type_update_job.rb
@@ -7,7 +7,12 @@ class SampleTypeUpdateJob < ApplicationJob
   # if refresh_samples is false, then the associated samples won't be refreshed, only the constraints cache rebuilt
   # - defaults to true
   def perform(sample_type, refresh_samples)
-    sample_type.refresh_samples if refresh_samples
+    if refresh_samples
+      sample_type.refresh_samples
+      sample_type.samples.each do |sample|
+        Rails.cache.delete_matched(/#{sample.list_item_title_cache_key_prefix}.*/)
+      end
+    end
     Seek::Samples::SampleTypeEditingConstraints.new(sample_type).refresh_cache
   end
 end
--- a/app/jobs/sample_type_update_job.rb
+++ b/app/jobs/sample_type_update_job.rb
@@ -9,10 +9,9 @@ class SampleTypeUpdateJob < ApplicationJob
   def perform(sample_type, refresh_samples)
     if refresh_samples
       sample_type.refresh_samples
-
-      Rails.cache.delete_matched(/#{sample_type.list_item_title_cache_key_prefix}.*/)
-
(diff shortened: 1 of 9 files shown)


app/jobs/sample_type_update_job.rb
```diff
@@ -7,7 +7,12 @@ class SampleTypeUpdateJob < ApplicationJob
   # if refresh_samples is false, then the associated samples won't be refreshed, only the constraints cache rebuilt
   # - defaults to true
   def perform(sample_type, refresh_samples)
-    sample_type.refresh_samples if refresh_samples
+    if refresh_samples
+      sample_type.refresh_samples
+      sample_type.samples.each do |sample|
+        Rails.cache.delete_matched(/#{sample.list_item_title_cache_key_prefix}.*/)
+      end
```
fbacall, line 14: Could make the cache key for sample RLI titles derived from the sample type's cache key, e.g.

`rli_title_sample_types/1/samples/123`

then you could have a single delete for ``rli_title_sample_types/1/samples/*` (or it might even by deleted automatically by the sample type's RLI title cache expiring)

stuzart: yes, good idea

stuzart: I have a few doubts now. 

Doing so would introduce a dependency and prior knowledge on how the cache key is generated, which isn't ideal but also not so bad, I can base it off the sample type key and make sure there are tests. 

But testing using 2000 samples, it only took 25 seconds, and it was actually quite hard to cache that many and required increasing the related items limit up to 2000. Without being previously cached it only took 1.5 second. With related items limit of 10 also took about 1.5 seconds.

So I'm not really sure it's worth it.

```
.7.5 :022 > puts Benchmark.measure { st.samples.each{|sample| Rails.cache.delete_matched(/#{sample.list_item_title_cache_key_prefix}.*/)} }
 17.954077   7.655812  25.609889 ( 25.665434)
 => nil 
2.7.5 :023 > puts Benchmark.measure { st.samples.each{|sample| Rails.cache.delete_matched(/#{sample.list_item_title_cache_key_prefix}.*/)} }
  0.922049   0.593479   1.515528 (  1.518624)
 => nil 
2.7.5 :024 > puts Benchmark.measure { st.samples.each{|sample| Rails.cache.delete_matched(/#{sample.list_item_title_cache_key_prefix}.*/)} }
  1.113349   0.451247   1.564596 (  1.565124)
 => nil 
```

stuzart: although if "(or it might even by deleted automatically by the sample type's RLI title cache expiring)" it would definitely be a win so will give it a go

stuzart: Updated, and also added fix for `.cache_key` change with Rails 6

Issue #1037, opened 2022-06-16
tagging teams and spaces, and not just with EDAM!


CaroleGoble:
there has been discussios  point about tagging everything (including teams and spaces) with open tags and EDAM tags. the limited discussion restricted to EDAM tags. But EDAM is a "Ontology of bioscientific data analysis and data management". 

WorkflowHub is NOT just about biology. It needs to cater for biodiversity, climate change, social sciences, physics, chemistry   - these are all projects that are starting to use it and moreover are contributing to its funding.  

(i) We cannot just restrict tagging to EDAM and (ii) we need tagging on everything - especially teams and spaces as right now we have to look at each Team individually to figure out what they do and we cannot filter on tags

stuzart:
we will address this initially by
[] - removing direct references to EDAM in the UI
[] - removing references to EDAM in the API, e.g. changing `edam_topics `to just `topics`
[] - double check that it is possible for a site admin to edit the Controlled Vocab, including adding new terms even if placeholders not yet from an ontology

we can already manually merge with additional ontologies, one by one or through a script, but I've raised #1044 to make this easier through the UI as it is likely to be a common requirement

PR #1116, opened 2022-07-25
Refactoring to remove EDAM specific references

stuzart:
Updated code, seeds and database keys etc to reflect that the annotations may not always be EDAM specific and to avoid future potential confusion. Now just refers to annotating via a controlled vocabulary.

Relates to #1037


app/helpers/assets_helper.rb
```diff
@@ -329,10 +329,10 @@ def request_contact_pending?(resource)
     ContactRequestMessageLog.recent_requests(current_user.try(:person), resource).first
   end
 
-  def edam_ontology_items(ontologyCVItems)
-    Array(ontologyCVItems).collect do |item|
-      browser_url = "https://edamontology.github.io/edam-browser/#{URI.parse(item.iri).path.gsub('/','#')}"
-      link_to(item.label, browser_url, target: :_blank).html_safe
+  def controlled_vocab_annotation_items(controlled_vocab_terms)
+    Array(controlled_vocab_terms).collect do |term|
+      browser_url = "https://edamontology.github.io/edam-browser/#{URI.parse(term.iri).path.gsub('/','#')}"
```
stuzart, line 334: I need to update this to unless show the link if edam is the source

PR #1117, opened 2022-07-25
Samples issues

vdkkia:
Querying samples via Templates has been added and some improvements have been made.


app/assets/javascripts/select2.full.min.js
```diff
@@ -1,2 +1,3 @@
-/*! Select2 4.1.0-rc.0 | https://github.com/select2/select2/blob/master/LICENSE.md */
```
stuzart, line 1: I think this should go in vendor/assets/javascripts/

app/models/sample.rb
```diff
@@ -130,6 +130,10 @@ def related_organism_ids
     organism_ids | ncbi_linked_organisms.map(&:id)
   end
 
+	def get_value(attribute)
```
stuzart, line 133: this isn't needed, should use `get_attribute_value(attr)`

app/views/sample_types/select.html.erb
```diff
@@ -3,11 +3,22 @@
...
+	<% if templates_enabled %>
+		<div class="col-md-12">
+			<%= link_to "Search by #{t('template').pluralize}", query_form_samples_path, class: "btn btn-default btn-lg" %>
+			<p style="font-size:20px;margin-top:20px">OR</p>
+		</div>
+	<% end %>
+  <div class="col-md-12">
+		<h1>Select <%=t('sample_type') %></h1>
+		<p>
+			Use the filters to find the <%=t('sample_type') %> you are looking for, and then click the New Sample button to create one of this type.
```
stuzart, line 18: will this work with `templates_enabled = false` ? Difficult to tell from the diff

app/helpers/samples_helper.rb
```diff
@@ -171,6 +171,17 @@ def ols_root_term_link(ols_id, term_uri)
...
   end
 
+  def get_extra_info(sample)
+    {
+      project_ids: sample.project_ids.join(','),
+      project_names: sample.projects.map { |p| link_to(p.title, p, target: :_blank) }.join(',').html_safe,
+      study_ids: sample.sample_type.study_ids.join(','),
+      study_names: sample.sample_type.studies.map { |s| link_to(s.title, s, target: :_blank) }.join(',').html_safe,
+      assay_ids: sample.sample_type.assay_ids.join(','),
+      assay_names: sample.sample_type.assays.map { |a| link_to(a.title, a, target: :_blank) }.join(',').html_safe
```
stuzart, line 181: what if the studies or assays are private ? Looks like the names will leak through

stuzart (changes requested):

app/assets/javascripts/select2.full.min.js
vdkkia: It was moved to vendor/assets/javascripts

app/views/sample_types/select.html.erb
vdkkia: It won't show the button if `templates_enabled = false`
Redirection from the `sample_qury` page added in case `templates_enabled = false`

app/helpers/samples_helper.rb
vdkkia: It's been fixed.

app/models/sample.rb
vdkkia: It's been fixed.

stuzart (approved):

Issue #1122, opened 2022-07-26
Ontology vs Controlled Vocabulary in Attribute Type


floradanna:
In DataHub, both Ontology and Controlled Vocabulary (CV) are shown as possible options to define the Attribute Type. 
However, the exact same list of CV is shown for both "Ontology" and "Controlled vocabulary", when selected. This doesn't make much sense; you would expect to see
- list of ontologies --> when selecting "Ontology"
- list of controlled vocabularies --> when selecting "Controlled vocabulary"

Moreover, users cannot use a CV list TITLE that has already been given to another CV list. 
@rabuono we should think about this since it prevents users to use the root ontology term as CV title more than just once.

floradanna:
@vdkkia how did you solve this? How are CV and Ontology listed now?

PR #1131, opened 2022-07-27
Fix Ontology and CV in Attribute Type issue #1122

vdkkia:
This filters out the CV and Ontology related items from the list based on the selected Sample Attribute Type


stuzart (approved):

stuzart:
Approved but it has some conflicts with work on the earlier PR

PR #1159, opened 2022-08-24
Single page improvements

vdkkia:
Some improvements have been made and a couple of bug fixes have been done.


app/models/investigation.rb
```diff
@@ -39,6 +43,10 @@ def related_publication_ids
     publication_ids | study_publication_ids | assay_publication_ids
   end
 
+  def related_sop_ids
```
stuzart, line 46: Do we have a unit test that checks this, and checks it combines the sops from the study and assay? I think there may be a cleaner way of doing this, but can be refactored in the future but a test would help with that refactoring.

stuzart (changes requested):

Issue #640, opened 2021-06-18
Add time zone information to Events


rabuono:
As a user I would like to be able to add time zone information when creating Events.
When not entering any time zone information, Event defaults to UTC.
Time zone would be recorded as difference to UTC.
Agreement needed on how to present time zone choices to user.

floradanna:
Could it be fixed as part of maintenance work?

stuzart:
> Could it be fixed as part of maintenance work?

No, the maintenance work is ending anyway. But we can plan to include this in the next set of tasks

alaninmcr:
As Stuart said, the maintenance sprint is coming to an end.

One solution could be would be to keep the time in UTC in the database, and to use Javascript to convert to the browser's local time zone when the event is displayed. Other calendar systems seem to do something similar.

When the event is created, you could specify the timezone and it would be converted to UTC for storage.

That's just a quick idea though.


________________________________________

fbacall:
@stuzart did you mean to move this from "Done" to "In progress"?

PR #1164, opened 2022-09-07
Timezone for events

vdkkia:
Add time zone information to Events #640

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -36,6 +36,7 @@ class EventsController < ApplicationController
     re_render_view = is_new ? 'events/new' : 'events/edit'
 
     update_sharing_policies @event
+    set_timezone
 
     respond_to do | format |
       if @event.update(event_params) && @event.save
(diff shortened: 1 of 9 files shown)


app/controllers/events_controller.rb
```diff
@@ -62,4 +63,15 @@ def param_converter_options
...
   end
 
+  def set_timezone
+    time_zone = params[:event][:time_zone]
+    if time_zone.present? && (ActiveSupport::TimeZone.all.map{ |t| t.tzinfo.name }.include? time_zone)
+      start_date = params[:event][:start_date]
+      end_date = params[:event][:end_date]
+
+      params[:event][:start_date] = start_date.in_time_zone(time_zone) if start_date.present?
+      params[:event][:end_date] = end_date.in_time_zone(time_zone) if end_date.present?
```
fbacall, line 73: Won't this mean the start/end times will shift each time the event is updated?

Since it will re-apply the timezone modification

app/models/event.rb
```diff
@@ -69,4 +71,15 @@ def time_zone_valid?
...
+    if time_zone.present? && time_zone_valid?
+      if start_date.present? && (start_date_changed? || time_zone_changed?)
+        self.start_date = start_date.to_s(:db).in_time_zone(time_zone)
+      end
+      if end_date.present? && (end_date_changed? || time_zone_changed?)
+        self.end_date = end_date.to_s(:db).in_time_zone(time_zone)
+      end
+    end
+  end
+
```
fbacall, line 84: Could we have some tests for this behaviour? :)

1. Check the times are shifted according to the time zone
2. Check the times do not shift on subsequent saves
3. Check the times DO shift if the timezone is changed

PR #1166, opened 2022-09-09
CFF parsing

fbacall:
Extract workflow metadata from `CITATION.cff` if present in the repo being registered


lib/seek/workflow_extractors/git_repo.rb
```diff
@@ -66,6 +69,14 @@ def abstract_cwl_extractor
         abstract_cwl_path = @git_version.path_for_key(:abstract_cwl)
         @abstract_cwl_extractor = abstract_cwl_path ? Seek::WorkflowExtractors::CWL.new(@git_version.file_contents(abstract_cwl_path, fetch_remote: true)) : nil
       end
+
+      def cff_extractor
+        return @cff_extractor if @cff_extractor
```
stuzart, line 74: could this be `... if defined?(@cff_extractor)` as it could be set to `nil `if there is no file, meaning it would repeatedly check for the file being present

stuzart (approved):
just something very minor

PR #1190, opened 2022-10-03
Single page improvements

vdkkia:


app/forms/isa_assay.rb
```diff
@@ -51,7 +51,7 @@ def populate(id)
 
   def validate_objects
     @assay.errors.each { |e| errors[:base] << "[Assay]: #{e.full_message}" } unless @assay.valid?
-    errors[:base] << '[SOP]: SOP is required' if @assay.sop_ids.blank?
+    # errors[:base] << '[SOP]: SOP is required' if @assay.sop_ids.blank?
```
stuzart, line 54: can we just delete this line ?

app/models/study.rb
```diff
@@ -70,7 +71,7 @@ def related_person_ids
   end
 
   def related_sop_ids
-    Array(sop_id) | assay_sop_ids
+    Array(sop_ids) | assay_sop_ids
```
stuzart, line 74: Array no longer needed, it should already be an array

app/views/isa_studies/_form.html.erb
```diff
@@ -47,12 +47,11 @@
...
     <% end %>
 
-		<%= folding_panel("SOP") do %>
-			<p class="help-block">The following SOP is associated with this study:</p>
-			<%= study_fields.collection_select(:sop, authorised_assets(Sop, current_user.person.projects, "view"),
-															:id, :title,
-															{prompt:'Select'},
-															{class:"form-control", id:'isa_study_sop_selector', name: "isa_study[study][sop_id]"}) %>
+		<%= folding_panel("SOPs") do %>
+			<p class="help-block">The following SOPs are associated with this study:</p>
```
stuzart, line 51: should be using the` t('sop').pluralize`

app/views/isa_studies/_sop.html.erb
```diff
@@ -1,28 +1,30 @@
...
+  <% end %>
 <% else %>
-	<p>
-		<em>
-			Warning: A Protocol is missing. Associate a Protocol before exporting ISA.
-		</em>
-	</p>
+  <p>
+    <em>
+      Warning: A Protocol is missing. Associate a Protocol before exporting ISA.
```
stuzart, line 27: says Protocol here, but elsewhere SOP ? Do they mean the same thing ?

stuzart (changes requested):
really minor changes

app/views/isa_studies/_sop.html.erb
floradanna: Yes. But we would like DataHub users to see/read Protocol instead of SOP (as IBISBA does). We will open a new issue for that.

PR #1208, opened 2022-10-19
Localizations for validations

abecam:
Together with the pop-up descriptions.

- All elements are localized (so in 3 places, 2 with aliases), not sure it's a good idea.
- Tests in test/unit/internationalization_test.rb are a bit stupid, but I thought they do the job (i.e. test the localization not something else).
- the list of links for help is directly in the config folder: config/help_links.yml


test/fixtures/mailer/gatekeeper_approval_feedback
```diff
@@ -1,7 +1,7 @@
 Hi Aaron Spiggle,
 
     Upon your publishing requests, the Sysmo SEEK -asset_gatekeeper- Gatekeeper Last, http://localhost:3000/people/-person_id- approved to publish:
-        Data file: Picture - http://localhost:3000/data_files/-df_id-
+        Datafile: Picture - http://localhost:3000/data_files/-df_id-
```
fbacall, line 4: Rather than change the fixture, can we change the mailer templates to use I18n?

(I guess they were relying on the inflections before)

test/unit/helpers/help_helper_test.rb
```diff
@@ -37,4 +39,31 @@ class HelpHelperTest < ActionView::TestCase
...
   end
+
+  test 'help icon' do
+    link = index_and_new_help_icon 'collection'
+    tag = Nokogiri::HTML::DocumentFragment.parse(link).children.first
+    
+    assert_equal 'https://docs.seek4science.org/help/user-guide/collections.html', tag['href']
+    assert_equal '_blank', tag['target']
+    assert_equal t("info_text.collection"), tag['data-tooltip']
+    assert_equal 'What is a Collection?', tag.children[1].content
```
fbacall, line 50: Would be good to also have a test that proves the `indefinite_article` method works for e.g. "an Assay".

fbacall (changes requested):
Looks good, just a few very minor things.

test/fixtures/mailer/gatekeeper_approval_feedback
abecam: Should be done. I use model_name.human as it should concern models only. But I am not super familiar with the gatekeeper so don't hesitate to correct me :)

test/unit/helpers/help_helper_test.rb
abecam: Really good idea and added :)

fbacall (approved):

PR #1222, opened 2022-10-28
Sample batch permission single page

vdkkia:


app/views/assets/sharing/batch_sharing_permission_preview.html.erb
```diff
@@ -1,83 +1,127 @@
 <%
+  single_page = params[:single_page]
+  sample_ids = (params[:sample_ids] || []).map(&:to_i)
+
```
stuzart, line 4: I think this and also `sharing_bulk_change_preview.html.erb` needs a rethink. There is just too much special `single_page` special behaviour, and sample specific stuff, mixed in with a view that is intended to be generic and polymorphic. 
This specific behaviour would better placed in their own views or partials

stuzart (changes requested):
there is much single_page or sample specific behaviour being embedded in generic views

PR #1223, opened 2022-10-31
Hidden Single Page Sample Types

vdkkia:


app/models/sample_type.rb
```diff
@@ -115,7 +117,13 @@ def can_delete?(user = User.current_user)
   end
 
   def can_view?(user = User.current_user, referring_sample = nil)
-    project_membership = (user && user.person && (user.person.projects & projects).any?)
+    return false if Seek::Config.project_single_page_advanced_enabled && !template_id.nil?
+
+    can_view_in_single_page?(user, referring_sample)
```
stuzart, line 122: I don't think this check is specific to the single page view ?

app/models/sample_type.rb
```diff
@@ -115,7 +117,13 @@ def can_delete?(user = User.current_user)
...
+  end
+
+  def can_view_in_single_page?(user = User.current_user, referring_sample = nil)
```
stuzart, line 125: As above, I feel the logic is the wrong way around here, as this method contains the default general behaviour, whereas can_view? contains logic specific to the single page view

stuzart (changes requested):

lib/treeview_builder.rb
```diff
@@ -37,7 +37,14 @@ def folder_node(folder)
   end
 
   def create_node(obj)
-    unless obj[:resource].can_view?
+    resource = obj[:resource].class.name.underscore
+    can_view = if resource == 'sample_type'
+                 obj[:resource].send(:can_view?, User.current_user, nil,
+                                     resource == 'sample_type')
```
stuzart, line 43: resource == 'sample_type' will always be true

lib/treeview_builder.rb
```diff
@@ -37,7 +37,14 @@ def folder_node(folder)
...
+               else
+                 obj[:resource].send(:can_view?)
```
stuzart, line 45: you can just call `can_view?` - no need to use send

lib/treeview_builder.rb
```diff
@@ -37,7 +37,14 @@ def folder_node(folder)
   end
 
   def create_node(obj)
-    unless obj[:resource].can_view?
+    resource = obj[:resource].class.name.underscore
+    can_view = if resource == 'sample_type'
```
stuzart, line 41: you can just do `obj[:resource].is_a?(SampleType)`

stuzart (changes requested):

PR #1224, opened 2022-10-31
Remove existing RO-Crate before attempting to write. Fixes #1221

### LINKED ISSUE #1221: RO-Crate generation fails with symlink in repo
error is:
```
       8: from app/models/concerns/workflow_extraction.rb:214:in `ro_crate_zip'
        7: from app/models/concerns/workflow_extraction.rb:209:in `ro_crate'
        6: from app/models/concerns/workflow_extraction.rb:209:in `tap'
        5: from app/models/concerns/workflow_extraction.rb:194:in `block in ro_crate'
        4: from lib/git/operations.rb:86:in `in_temp_dir'
        3: from lib/git/operations.rb:88:in `block in in_temp_dir'
        2: from app/models/concerns/workflow_extraction.rb:196:in `block (2 levels) in ro_crate'
        1: from app/models/concerns/workflow_extraction.rb:215:in `block in ro_crate_zip'
Zip::EntryExistsError (add failed. Entry utils/make_csv/make_csv.py already exists)

```

repo/symlink: https://github.com/emo-bon/pipeline-v5/tree/eosc-life-gos/utils/make_csv

In theory it should handle symlinks since:
https://github.com/ResearchObject/ro-crate-ruby/commit/545f3f098d0636d9aaa78965778ecb330cdbc66a
https://github.com/rubyzip/rubyzip/pull/531

fbacall:


app/models/concerns/workflow_extraction.rb
```diff
@@ -212,7 +212,9 @@ def ro_crate
 
   def ro_crate_zip
     ro_crate do |crate|
-      ROCrate::Writer.new(crate).write_zip(ro_crate_path)
+      path = ro_crate_path
+      File.unlink(path) if File.exist?(path)
```
stuzart, line 216: why `unlink` rather than `delete`, which I think would be clearer. I'm not sure of the difference between them

stuzart (changes requested):
more a curiosity than a particular problem

app/models/concerns/workflow_extraction.rb
fbacall: There's no difference I don't think. I think I was using `unlink` in another code base so it came to mind, but it seems SEEK already makes use of `File.delete` elsewhere so I'll change it to be consistent

PR #1253, opened 2022-11-14
Last updated by

fherreazcue:
Added link to user who performed the last update inside the activity box, with its corresponding unit and functional tests.

### DIFF
--- a/app/views/assets/_usage_info_box.html.erb
+++ b/app/views/assets/_usage_info_box.html.erb
@@ -22,9 +22,9 @@
       <p>
         <strong>Last updated</strong>: <%= date_as_string(resource.updated_at, true) -%>
       </p>
-      <% unless ActivityLog.where(activity_loggable:resource, action:'update').last.culprit.nil? -%>
+      <% unless resource.updated_last_by.nil? -%>
         <p>
-          <strong>Last updated by</strong>: <%= ActivityLog.where(activity_loggable:resource, action:'update').last.culprit.person.name -%>
+          <strong>Last updated by</strong>: <%= resource.updated_last_by.person.name -%>
         </p>
       <% end -%>
     <% end -%>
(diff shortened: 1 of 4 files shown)


test/functional/assays_controller_test.rb
```diff
@@ -1860,4 +1860,62 @@ def check_fixtures_for_authorization_of_sops_and_datafiles_links
...
+    a = Factory :assay, policy: Factory(:public_policy), created_at: 15.minute.ago, contributor: person1
+    Factory :activity_log, activity_loggable: a, action: 'create', created_at: 15.minute.ago, culprit: person1
+    login_as(person1)
+    get :show, params: { id: a.id }
+    assert_response :success
+    assert_select 'span.updated_last_by', false, 'Last editor should not be shown just after creation'
+    Factory :activity_log, activity_loggable: a, action: 'update', created_at: 10.minute.ago, culprit: person1
+    get :show, params: { id: a.id }
+    assert_response :success
+    assert_select 'span.updated_last_by', person1.name
```
stuzart, line 1875: this would be better if the css selector was `span.updated_last_by a`, and the check was `text:person1.name`

you should also check the link, with `span.updated_last_by a[href=?]` and check for `person_path(person1)`

app/views/assets/_usage_info_box.html.erb
```diff
@@ -22,6 +22,11 @@
       <p>
         <strong>Last updated</strong>: <%= date_as_string(resource.updated_at, true) -%>
       </p>
+      <% unless resource.updated_last_by.nil? || !current_user&.person&.member_of?(resource.projects)-%>
```
stuzart, line 25: instead, this would be better as

`if resource.updated_last_by && current_user&.person&.member_of?(resource.projects)`

avoiding double negatives

stuzart: `nil` can be treated equivalent as false in Ruby

stuzart (changes requested):

Issue #1198, opened 2022-10-12
Display Sample Type creator/submitter and creation date


rabuono:
As a User I would like to see who created a Sample Type and when it was created so that I can contact this person to: ask for instructions, ask for changes to the Sample Type.

Outcome of Samples WG and part of discussion #1118

stuzart:
I think this should be quite straight forward, so added to the 1.13.1 milestone (it might even make 1.13.0)

rabuono:
Add the ability to define independent Creator and Submitter to a Sample Type

stuzart:
Following testing this at https://testing.sysmo-db.org - it works but some additional refinements are needed:

- [ ] Include filtering by submitter on the index page
- [ ] Display the submitter icon on the resource list view ( e.g. avatar on the left of the items https://testing.sysmo-db.org/sops )
- [ ] Upgrade task to set the creator to the submitter for sample types before this change

PR #1256, opened 2022-11-17
Sample type info

fherreazcue:
Adds main content right panel to sample types show.


stuzart (approved):

stuzart:
I will leave the issue #1198  open for now whilst we get more details about handling creators

PR #1271, opened 2022-11-24
Minor LifeMonitor fixes

fbacall:
Stops some submission jobs being created when the daily test status checker job is run.

Also makes `scopes` configurable in the client.


fbacall:
Don't merge yet, I found another bug

Issue #1288, opened 2022-12-07
Error when deleting a project as project admin


stuzart:
Although you get the option to, when deleting the project (with no items other than members) you get an error "Admin rights required"

This should have been updated with #383

PR #1294, opened 2022-12-09
fix to allow project administrator to delete project #1288

stuzart:
Fix to #1288 - incorrect filter was preventing a non-admin from deleting. Also made sure other actions were authorized correctly.


stuzart:
spotted there are overlapping callbacks that can be refactored slightly, so closing and will redo

PR #1295, opened 2022-12-09
Snapshots

fherreazcue:
Added custom title and description to snapshots, and the option to edit them as long as they don't have a DOI assigned.

### DIFF
--- a/lib/zenodo/acts_as_zenodo_depositable.rb
+++ b/lib/zenodo/acts_as_zenodo_depositable.rb
@@ -42,10 +42,6 @@ module Zenodo
         metadata = zenodo_metadata
         metadata.merge!(extra_metadata.to_h.deep_symbolize_keys)
 
-        #FIXME: this is a quick hack
-        metadata[:title] = 'not set' if metadata[:title].blank?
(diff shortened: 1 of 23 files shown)


stuzart:
Paused until the new year, need to look more closely at how metadata is updated and passed

fherreazcue:
Merged main to branch, updated tests and resolved conflicts.

fbacall:
Is this for 1.14 or 1.15?

lib/zenodo/acts_as_zenodo_depositable.rb
```diff
@@ -43,6 +43,7 @@ def export_to_zenodo(access_token, extra_metadata = {})
         metadata.merge!(extra_metadata.to_h.deep_symbolize_keys)
 
         #FIXME: this is a quick hack
+        metadata[:title] = 'not set' if metadata[:title].blank?
```
fbacall, line 46: The Zenodo export needs thinking about, because the snapshot title/description may not be what you want to be the title/description of the Zenodo entry (currently it would be the title/description of the resource being snapshotted).

PR #1308, opened 2022-12-15
Cookie consent

abecam:
Complete with tests, but might not be the most elegant everywhere.


app/helpers/assets_helper.rb
```diff
@@ -38,8 +38,15 @@ def asset_field_prefix
   # will render a view of the asset, if available. For example, a slideshare based asset could give a embedded slideshare view
   def rendered_asset_view(asset)
     return '' unless asset.can_download?
-    content = Rails.cache.fetch("#{asset.cache_key}/#{asset.content_blob.cache_key}") do
-      Seek::Renderers::RendererFactory.instance.renderer(asset.content_blob).render
+
+    ourRenderer = Seek::Renderers::RendererFactory.instance.renderer(asset.content_blob)
```
fbacall, line 42: Should be snake case

app/controllers/concerns/raw_display.rb
```diff
@@ -11,8 +11,11 @@ def render_display(blob, url_options: {})
       renderer = Seek::Renderers::RendererFactory.instance.renderer(blob, url_options: url_options)
     end
     if renderer.can_render?
-      response.set_header('Content-Security-Policy', renderer.content_security_policy)
-      render renderer.format => renderer.render_standalone.html_safe, layout: renderer.layout
+      # check if allowed by cookies
+      unless renderer.is_remote? && !cookie_consent.allow_embedding?
+        response.set_header('Content-Security-Policy', renderer.content_security_policy)
+        render renderer.format => renderer.render_standalone.html_safe, layout: renderer.layout
+      end
```
fbacall, line 18: What does this actually render if embedding is forbidden?

Would be nice if it displayed some kind of page with a message.

app/views/admin/features_enabled.html.erb
```diff
@@ -117,7 +117,12 @@
   <div id="add_default_templates" class="additional_settings" style="<%= show_or_hide_block Seek::Config.sample_type_template_enabled -%>">
 		<%= link_to("Add default #{t(:template).pluralize}", default_templates_templates_path, class: 'btn btn-default', id: 'add_default_templates_btn', disabled: !Seek::Config.sample_type_template_enabled) %>
 	</div>
-                             
+
+  <%= admin_checkbox_setting(:require_cookie_consent, 1, Seek::Config.require_cookie_consent,
+                             "Add a cookie consent banner and a cookie setting page", "Show a banner asking for consent for the use of cookies. Seek uses an essential cookie for session handling, and optional cookies for tracking if set below.
```
fbacall, line 122: Seek -> SEEK

app/views/cookies/consent.html.erb
```diff
@@ -0,0 +1,47 @@
+<div class="row">
+  <div class="col-md-8">
+    <h1>Cookie Consent</h1>
+
+    <%= render partial: 'general/cookies_info' %>
+
+    <% analytics_txt = "Google Analytics" if (Seek::Config.google_analytics_enabled && !Seek::Config.google_analytics_enabled) %>
+    <% analytics_txt = "Matomo" if (Seek::Config.piwik_analytics_enabled && !Seek::Config.google_analytics_enabled) %>
+    <% analytics_txt = "Google Analytics and Matomo" if (Seek::Config.piwik_analytics_enabled && Seek::Config.google_analytics_enabled) %>
+    <% analytics_txt = "Google Analytics" if (Seek::Config.google_analytics_enabled) %>
```
fbacall, line 10: The first option can never be satisfied 

Could probably simplify this by adding the strings to an array and calling `.join(' and ')`

lib/seek/renderers/blob_renderer.rb
```diff
@@ -26,6 +26,10 @@ def can_render?
         fail 'needs to be implemented'
       end
 
+      def is_remote?
+        false
+      end
```
fbacall, line 31: I wonder if this needs a better name to explain what it is used for... maybe `external_embed?` or something

app/helpers/assets_helper.rb
abecam: Changed, Java habit... :(

app/views/admin/features_enabled.html.erb
abecam: Changed, also in en.yml

app/views/cookies/consent.html.erb
abecam: Did that, but also had to change 2 different parts below (tracking shouldn't be shown if not activated, the last lines were redundant and kind of stupid). Not sure there isn't a better way to show the selected choices.

lib/seek/renderers/blob_renderer.rb
abecam: Changed :)

stuzart (approved):

PR #1331, opened 2023-01-31
Seek 1.13 add cvlist

whomingbird:


lib/seek/samples/attribute_type_handlers/cv_list_attribute_type_handler.rb
```diff
@@ -0,0 +1,29 @@
+module Seek
+  module Samples
+    module AttributeTypeHandlers
+      class CVListAttributeTypeHandler < BaseAttributeHandler
+        class MissingControlledVocabularyException < AttributeHandlerException; end
```
stuzart, line 5: this exception already exists for `CVAttributeTypeHandler`. Infact I noticed another duplicate of `MissingLinkedSampleTypeException` (not related to this PR).

Not a blocker for this PR but perhaps a future refactoring task to bring them all under `AttributeTypeHandlers::`

lib/seek/samples/attribute_type_handlers/cv_list_attribute_type_handler.rb
```diff
@@ -0,0 +1,29 @@
...
+        class NotArrayException < AttributeHandlerException; end
+        class InvalidControlledVocabularyException < AttributeHandlerException; end
+
+        def test_value(array_value)
+
+          if !array_value.is_a?(Array)
+            fail NotArrayException.new unless array_value.is_a?(Array)
+          else
+            array_value.each do |value|
+              fail InvalidControlledVocabularyException.new(value) unless controlled_vocab.includes_term?(value) || controlled_vocab.custom_input
```
stuzart, line 15: if you provide a term that isn't in the controlled vocab, won't this result in an unhandled exception rather than validation being false?
There exceptions are passed forward as something unexpected happening, rather than just a validation error - https://github.com/seek4science/seek/blob/master/lib/seek/samples/attribute_type_handlers/base_attribute_handler.rb#L23

stuzart, line 15: if you provide a term that isn't in the controlled vocab, won't this result in an unhandled exception rather than validation being false?
There exceptions are passed forward as something unexpected happening, rather than just a validation error - https://github.com/seek4science/seek/blob/master/lib/seek/samples/attribute_type_handlers/base_attribute_handler.rb#L23

test/unit/custom_metadata_attribute_test.rb
```diff
@@ -62,6 +62,22 @@ class CustomMetadataAttributeTest < ActiveSupport::TestCase
...
+    assert attribute.validate_value?(nil)
+    assert attribute.validate_value?('')
+    assert attribute.validate_value?([])
+    assert attribute.validate_value?(['Granny Smith'])
+
+    assert_raises(Seek::Samples::AttributeTypeHandlers::CVListAttributeTypeHandler::NotArrayException) do
+      assert attribute.validate_value?('Granny Smith')
+    end
+
+    assert_raises(Seek::Samples::AttributeTypeHandlers::CVListAttributeTypeHandler::InvalidControlledVocabularyException) do
```
stuzart, line 78: it shouldn't be raising an exception just for a validation error

stuzart (changes requested):

test/unit/custom_metadata_attribute_test.rb
stuzart: I mean this type of Exception, that is based on AttributeHandlerException

stuzart (approved):

PR #1346, opened 2023-02-08
Adds options to sort by downloads or views

fherreazcue:
Resolves issue #1098

### DIFF
--- a/lib/seek/list_sorter.rb
+++ b/lib/seek/list_sorter.rb
@@ -56,6 +56,17 @@ module Seek
                          arel_field = items.arel_table[:downloads].desc
                          arel_field
 
+                         # Alternative query construction:
+                         # alog = ActivityLog.all.arel_table
+                         # joined=items.joins(
+                         #     Arel::Nodes::OuterJoin.new(alog, alog.create_on(
+                         #       alog[:activity_loggable_id].eq(items.arel_table[:id])
+                         #         .and(alog[:action].eq('download'))
+                         #         .and(alog[:activity_loggable_type].eq('DataFile'))
+                         #     )))
+                         #   .select(:id, :title, alog[:activity_loggable_id].count.as('downloads'))
+                         #   .group(:id,:title)
+
                          # Play in the console with:
                          # items=DataFile.all
                          # items.object_id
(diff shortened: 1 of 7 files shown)


lib/seek/list_sorter.rb
```diff
@@ -38,6 +39,82 @@ class ListSorter
...
+                          #### Using Arel
+                          # This section builds the equivalent arel_table to provide the corresponding arel_field
+                          items_a=items.arel_table
+                          alog_a=ActivityLog.arel_table
+                          downloads=alog_a.project(alog_a[:activity_loggable_id].as("log_id"), alog_a[:activity_loggable_id].count.as("downloads"))
+                                        .where(alog_a[:action].eq('download').and(alog_a[:activity_loggable_type].eq(items.first.class.name)))
+                                        .group(:activity_loggable_id).as('downloads')
+                          joined=items_a.project(items_a[Arel.star], downloads[:downloads]).outer_join(downloads).on(items_a[:id].eq(downloads[:log_id])).as('d')
+
+                          case ActiveRecord::Base.connection.instance_values["config"][:adapter]
```
stuzart, line 59: This all seems very complex and long, and will be difficult to revisit and maintain in the future. Needs simplifying or breaking down. It could also be inconsistent with the `download_counts`, as it isn't include the `no_spider ` scope ( which needs improving, but will be more of a reason to include when it has been).

I wonder if there might be a way of changing the `count_actions ` queries in https://github.com/seek4science/seek/blob/master/lib/seek/stats/activity_counts.rb to a scope, and then chaining them together.

A possibility may be to split the query up rather than do one big join.

Another option may be to just sort programmatically, and then turn back into a relation (as is done with --relevence), doing this with all Datafiles on the FAIRDOMHub takes about 30 seconds but that could be an extreme case

lib/seek/list_sorter.rb
```diff
@@ -38,6 +39,82 @@ class ListSorter
...
+                      #### Using Active Record
+                      # This section **modifies** "items" relation so that it includes a new column "views"
+                      alog=ActivityLog.all.where(action: 'show',activity_loggable_type: items.first.class.name)
+                      views=alog.select("activity_loggable_id AS #{items.table.name}_id, COUNT(activity_loggable_id) AS views").group(:activity_loggable_id)
+                      items._select!('*', 'v.views').joins!("LEFT OUTER JOIN (#{views.to_sql}) v ON #{items.table.name}.id = v.#{items.table.name}_id")
+
+                      #### Using Arel
+                      # This section builds the equivalent arel_table to provide the corresponding arel_field
+                      items_a=items.arel_table
+                      alog_a=ActivityLog.arel_table
```
stuzart, line 91: same as above

stuzart (changes requested):

stuzart (approved):
refactoring can be done at a later stage if necessary

PR #1350, opened 2023-02-09
Makes studies and investigations taggable.

### LINKED ISSUE #1327: be able to  also "tag"  Investigation and Study
image

add the "tags" block also for the page to create/update Investigation and Study

fherreazcue:
Resolves #1327 .

### DIFF
--- a/app/models/investigation.rb
+++ b/app/models/investigation.rb
@@ -1,6 +1,4 @@
 class Investigation < ApplicationRecord
-  include Seek::Rdf::RdfGeneration
-  include Seek::Taggable
 
   acts_as_isa
   acts_as_snapshottable
--- a/app/models/investigation.rb
+++ b/app/models/investigation.rb
@@ -1,5 +1,6 @@
 class Investigation < ApplicationRecord
   include Seek::Rdf::RdfGeneration
+  include Seek::Taggable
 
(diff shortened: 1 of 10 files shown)


app/models/investigation.rb
```diff
@@ -1,5 +1,6 @@
 class Investigation < ApplicationRecord
   include Seek::Rdf::RdfGeneration
+  include Seek::Taggable
```
stuzart, line 3: since all Investigation, Study and Assay now supports tagging, this can be moved into `acts_as_isa` ( same for `Study`)

stuzart: incidentally, the same could be said for the existing `include Seek::Rdf::RdfGeneration` and that could be moved too.

stuzart (approved):
Just a minor suggestion. Once that's updated ( or there is a problem doing so ) please feel free to merge and close

app/models/investigation.rb
fherreazcue: Resolved in [c77b9c6](https://github.com/seek4science/seek/pull/1350/commits/c77b9c6ab7da4bf566374f5aed70470f56ac28c9).
Moved both RdfGeneration and Taggable to act_as_isa.

PR #1362, opened 2023-02-16
Ldap truncate username

abecam:
A very small change to truncate usernames coming from OAuth so they don't go above the size limit(40) and don't clash (so allow the method to avoid clashes to work).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -294,7 +294,11 @@ class User < ApplicationRecord
 
   def self.from_omniauth(auth)
     User.new.tap do |user|
-      user.login = unique_login(auth['info']['nickname'] || 'user')
+      username_from_oauth = (auth['info']['nickname'] || 'user').truncate(30)
+      # username are limited to 40, so we truncate it, with some extra-space as
+      # unique_login might add 4 numbers to avoid clashes
+
+      user.login = unique_login(username_from_oauth)
       user.password = random_password
       user.password_confirmation = user.password
     end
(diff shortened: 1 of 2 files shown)


stuzart:
would removing or increasing the limit on the User.login also solve the problem?

stuzart:
> > would removing or increasing the limit on the User.login also solve the problem?
> 
> Yes, the issue we have is that our identity server provide a kind of long hash for the uid, so >40 chars (I believe around 70). The truncation has the benefit of not worrying about the maximum size anymore. Feel free to reject the PR :) !

So, if we increased the limit to 120 (to avoid very long usernames), that would make this PR redundant, or is there still potential for name clashes ?

abecam:
> > > would removing or increasing the limit on the User.login also solve the problem?
> > 
> > 
> > Yes, the issue we have is that our identity server provide a kind of long hash for the uid, so >40 chars (I believe around 70). The truncation has the benefit of not worrying about the maximum size anymore. Feel free to reject the PR :) !
> 
> So, if we increased the limit to 120 (to avoid very long usernames), that would make this PR redundant, or is there still potential for name clashes ?

The unique_name method should avoid any name clashes anyway. Otherwise it would be difficult to be 100% sure as it depends on the OAuth provider.

stuzart:
> > > > would removing or increasing the limit on the User.login also solve the problem?
> > > 
> > > 
> > > Yes, the issue we have is that our identity server provide a kind of long hash for the uid, so >40 chars (I believe around 70). The truncation has the benefit of not worrying about the maximum size anymore. Feel free to reject the PR :) !
> > 
> > 
> > So, if we increased the limit to 120 (to avoid very long usernames), that would make this PR redundant, or is there still potential for name clashes ?
> 
> The unique_name method should avoid any name clashes anyway. Otherwise it would be difficult to be 100% sure as it depends on the OAuth provider.

I've created a PR #1369 - which if you approve and believes things, then you can close the PR

PR #1367, opened 2023-02-17
Extract bio.tools IDs from Galaxy workflows

fbacall:
When a user registers a Galaxy workflow, automatically extract bio.tools identifiers.

SEEK will create and store a mapping of Galaxy tool ID -> bio.tools IDs using the APIs of a configured set of Galaxy instances, and the bio.tools API (to fetch the tool names).


lib/galaxy/tool_map.rb
```diff
@@ -0,0 +1,87 @@
+module Galaxy
+  class ToolMap
+    CACHE_KEY = 'galaxy-bio-tools-map'.freeze
+
+    def self.instance
```
stuzart, line 5: I found this a bit confusing and unclear, as it appears to be using the Singleton pattern, but in a slightly weird way. It can be used either via the `.instance` method or through the public initializer. However, `.new` is only used in some tests, but not how it is generally used. Some instance methods are also public, but should probably be private.

I think it would clearer and simpler if it using the ruby `Singleton` module, and internally just cached `@map`

stuzart (changes requested):

PR #1369, opened 2023-02-20
Increase max login length to 120

### LINKED ISSUE #1366: Increase max size of login name
Currently set to 40, will increase to 120 to give more flexibility to still avoid very long usernames

stuzart:
Increase the login username limit, fixes #1366 and should avoid the need for #1362

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -34,7 +34,7 @@ class User < ApplicationRecord
   validates     :password_confirmation, presence: true, if: :password_required?
   validates_length_of       :password, minimum: MIN_PASSWORD_LENGTH, if: :password_required?
   validates_confirmation_of :password, if: :password_required?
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -34,7 +34,7 @@ class User < ApplicationRecord
   validates     :password_confirmation, presence: true, if: :password_required?
   validates_length_of       :password, minimum: MIN_PASSWORD_LENGTH, if: :password_required?
   validates_confirmation_of :password, if: :password_required?
-  validates_length_of       :login, within: 3..40
+  validates_length_of       :login, within: 3..120
```
fbacall, line 37: Maybe move this into a constant like `MIN_PASSWORD_LENGTH`?

abecam:
All fine for me, also tested it with our identity server with LDAP, working without issue!

Issue #1371, opened 2023-02-21
Update tagging and use of the typeahead UI to use select2 instead


stuzart:
select2 has been used by contributors, and has been agreed to be preferred - however, this has lead to inconsistencies in the UI. Decision is to switch to select2 throughout by updating and refactoring `object_input` and associated javascript and query methods.

Need to be careful about API breaking changes, where values were previously a comma separated list but now instead be an Array.

Relates to and motivated by #1334

Issue #1334, opened 2023-02-03
add a new sample attribute type, which could store an array of controlled vocabularies  


whomingbird:
This new sample attribute type can be used to create the sample type and the custom metadata attribute type: 

Compared to the "Controlled Vocabulary" sample attribute type,  the new type "Controlled Vocabulary List" is used to store an array of controlled vocabularies (strings).  see the image below : 

![image](https://user-images.githubusercontent.com/42063448/216615052-52e60765-c793-4b6d-b7a3-688ffad8149b.png)

whomingbird:
#1331

stuzart:
The new multi item autocomplete box needs applying to other places for consistency

PR #1372, opened 2023-02-21
Update CVList to use object_input and the typeahead component

stuzart:
Intermediate step before addressing #1371, so there is a releasable version of #1334 whilst that is addressed and ties it down to the `object_input` helper.

Temporarily requires the values to be passed and handled as a comma separated list, but this will be resolved when switched back to `select2` .

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -91,7 +91,7 @@ module SamplesHelper
       when Seek::Samples::BaseType::CV
         seek_cv_attribute_display(value, attribute)
       when Seek::Samples::BaseType::CV_LIST
-        value.join(", ")
+        value.each{|v| seek_cv_attribute_display(v, attribute) }.join(', ')
       else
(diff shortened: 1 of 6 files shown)


whomingbird (approved):

app/helpers/samples_helper.rb
```diff
@@ -30,21 +30,17 @@ def controlled_vocab_form_field(sample_controlled_vocab, element_name, values, l
...
-      select_tag element_name,
-                 options,
-                 class: "form-control cv_list_attribute",
-                 include_blank: "",
-                 name: "#{element_name}[]",
-                 multiple: "multiple"
+    scv_id = sample_controlled_vocab.id
+    is_ontology = sample_controlled_vocab.source_ontology.present?
+    existing_objects = Array(values).collect do |value|
+      Struct.new(:id, :name).new(value, value)
```
fbacall, line 38: The `Struct` should be moved outside the loop (and maybe put in a constant), or it will create a new type of Struct for each iteration

Issue #1382, opened 2023-03-08
Datafile attributions not set when using wizard


stuzart:
can be set through the usual form, but not via the wizard. Parameter may not be being passed through and handled correctly.
Would be solved by #1347

Issue #1381, opened 2023-03-08
Revisit custom_input flag for SampleControlledVocab


stuzart:
`custom_input` flag allows the user to add custom values, and not locked and validated against the controlled vocab.

currently it sets a `data-is-ontology` attribute on the field, and is then hidden from the available options by javascript

it's not necessarily an ontology based SCV, and am not sure why it is hidden ? (@floradanna / @rabuono ??)

it would also be more appropriate and flexible to apply this flag to the sample attribute where the SCV is used, so that the same SCV can but used with or without custom inputs depending on how the attribute is used.

floradanna:
is it related to #1147 ?

stuzart:
> is it related to #1147 ?

no

stuzart:
initially I will remove `data-is-ontology` and the associated javascript to hide it, and just filter out items from the select field that have `.custom_input?` as true. For datahub the behaviour will be the same.

floradanna:
@kdp-cloud @stuzart is this issue resolved? or not yet?

stuzart:
> @kdp-cloud @stuzart is this issue resolved? or not yet?

yes it has, it was fixed in a duplicate issue #1648 and will be part of 1.15.0. Feel free to close the issue

PR #1383, opened 2023-03-08
Tagging update to use select2 

### LINKED ISSUE #1371: Update tagging and use of the typeahead UI to use select2 instead

stuzart:
resolves #1371 

Replacing the old bootstrap typeahead with the select2 - for autocomplete selection

A knock on effect was extending api tests and examples to include multi samples and cv lists, which led to a need to update the schema as sample attribute maps properties need to support arrays.
Also extending tests, examples and api spec for data file data type and format annotations, to improve test coverage

Some possible next steps and observations, out of scope for this task but will be raised as new issues

- it's likely we can now remove bootstrap typeahead from the dependencies entirely
- SampleControlledVocabs can have a custom_input flag, to say that it allows new items. However, in the markup it is flagged as data-is-ontology, but may not be an ontology, and is then hidden from the selection with javascript
	- need to check why it is hidden and update the property
	- would be better to make this an option on the attribute, rather than the SCV
	- #1381 
- Use of select2 for Publications has been left untouched, as this doesn't use the standard tagging helpers
- API for sample attributes that are a list accept either comma separated or as an array - would be good to lock this down to an array and document
- Admin tag management is a bit weird, and includes funding codes. You can edit a tag, but not independently of it's type (e.g. would edit a tag for expertise, tools, tagging all in one go)
- #1382 bug discovered


app/assets/javascripts/objects_input.js
```diff
@@ -1,61 +1,54 @@
-function loadObjectInputs(elem) {
```
fbacall, line 1: This function is still used in various places to enable the tagging for elements that are added to the page dynamically

https://github.com/seek4science/seek/blob/902167afe1405cbad6bc8cdee7553bfc504840de/app/views/data_files/create_content_blob.js.erb#L30

app/assets/javascripts/objects_input.js
```diff
@@ -1,61 +1,66 @@
...
+$j(document).ready(function () {
 
-$j(document).ready(loadObjectInputs);
+    ObjectsInput.init();
+
+    $j('[data-role="seek-suggested-tags"]').on('click', function () {
+        let selectName = $j(this).data('tag-input');
+        let text = $j(this).text();
+        let selector = $j('select#'+selectName);
+        let vals = selector.val() || [];
```
fbacall, line 60: Very minor thing but these should probably be `const` instead of `let`

fbacall (approved):
One pedantic comment, but otherwise looks good

PR #1386, opened 2023-03-09
add read API for custom metadata 

whomingbird:


app/serializers/base_serializer.rb
```diff
@@ -121,6 +121,12 @@ def BaseSerializer.permits policy
     end
   end
 
+  attribute :custom_metadata_attributes, if: -> { object.respond_to?(:custom_metadata) && !object.custom_metadata.blank? } do
```
stuzart, line 124: As we're planning on renaming the code to extended metadata, at least for future proofing this attribute should be called `extended_attributes` and likewise the `extended_metadata_type_id` below. I think these are the only cases with the custom metadata naming is exposed

stuzart (changes requested):
Minor change. 
Needs tests and docs updated, but that can be discussed at a later stage

stuzart (approved):

Issue #1337, opened 2023-02-06
Empty Programme does not appear in My Items


rabuono:
**Describe the bug**
When clicking `My Items` an empty Programme to which the User is an admin does not appear.

**To Reproduce**
1. Create a Programme
2. Create a Project inside the Programme
3. Delete the Project inside of the Programme, thus making the Programme empty
4. The Programme now does not appear in `My Items`, it displays 

**Expected behavior**
A Programme, even if without Projects, should appear in `My Items` of its administrator. The Programme, even if empty, should also be listed under the `Related items` for its admin.

**Screenshots**
Bob is Programme admin
![image](https://user-images.githubusercontent.com/77321541/217023294-43caef0b-1e6c-4186-b429-f5ab39973b27.png)

The Programme is empty
![image](https://user-images.githubusercontent.com/77321541/217023474-fa2bf1b6-936b-480e-912a-b3a691ebafe8.png)

The Programme does not show in `My Items` for Bob
![image](https://user-images.githubusercontent.com/77321541/217023588-35c3cc45-f1f2-4f20-b757-de61c603093c.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox 109.0.1
 - Version SEEK 1.13.0

**Additional context**

fbacall:
@kdp-cloud following the tech call I looked at the branch, and I would go about this in a different way.

I don't think any changes to the database are needed, SEEK is already storing the link between the person and the programme via the programme administrator role, it just needs some tweaks to expose this relationship.

Check out the "related items" code which is used to populate lists of items related to a resource (and I assume it is used for "My Items"):
https://github.com/seek4science/seek/blob/master/lib/seek/related_items.rb
it's a bit metaprogram-y, but basically it tries to find the correct method through which to get related items to a given resource.

sometimes we implement bespoke methods to get related items where the relationship isn't quite straightforward:
https://github.com/seek4science/seek/blob/master/lib/seek/acts_as_isa/relationships.rb#L6-L18
https://github.com/seek4science/seek/blob/master/app/models/study.rb#L62-L73

This recent commit did something similar:
https://github.com/seek4science/seek/commit/c658564ff8b64917dc4fe44fd4b03968e60edc2d#diff-afb6f8bc3bae71b75743e00882a060863e2430cbe858ec9014e5956504dfc61cR61

kdp-cloud:
@fbacall  Thanks for looking into this!
[...]

PR #1400, opened 2023-03-17
Seek 1.13 issue 1337 show empty programmes

kdp-cloud:
This PR addresses [issue 1337](https://github.com/seek4science/seek/issues/1337).
The programme administrators were added as a related resource so empty programmes would show in the "my items" page


app/models/person.rb
```diff
@@ -129,6 +129,27 @@ def person
...
+
+  # Returns the ids of the programmes related to this person's projects.
+  def programme_ids_related_to_projects
+    Project.includes(:people).where(people: {id: self.id}).map(& :programme_id)
+  end
+
+  # Gets all programme id's who has an programme administrator id equal to this id.
+  def administrator_programme_ids
+    Programme.includes(:programme_administrators).where(programme_administrators: {id: self.id}).map(& :id)
+  end
```
stuzart, line 151: you don't actually need all this. you just need `def related_programmes` which returns

``` programmes || programmes_for_role(:programme_administrator) ``` 

instead of `programmes_for_role`,  there is also the cleaner `person.administered_programmes`  but this will also return all programmes if the person is an administrator, so you might want to refactor that slightly

**Review by stuzart — CHANGES_REQUESTED** [review]
also needs unit and functional tests

File: app/models/person.rb (hunk 1 above)
Comment (stuzart) [reply app/models/person.rb:151]: https://github.com/seek4science/seek/blob/eb1fc7e2cea43763e951fa831894a419636935ff/lib/seek/roles/accessors.rb#L75

Comment (stuzart) [reply app/models/person.rb:151]: I realised I made an error, that was meant to be a single `|` , i.e.
```
programmes | programmes_for_role(:programme_administrator)
```

kdp-cloud: `programmes | programmes_for_role(:programme_administrator)` does the trick and is indeed a lot cleaner.
Thanks!

I've also pushed a unit test and a functional test.

stuzart (approved):

Issue #1394, opened 2023-03-13
Make sure Submitter is added as Creator by default in Single Page


floradanna:
Make sure Submitter is added as Creator by default in Single Page.

When creating via Single Page:
- Investigation doesn't have Creator by default.

floradanna:
Related to #1390

PR #1402, opened 2023-03-23
Issue 1394 make submitter default creator in single page

kdp-cloud:
This PR addresses the bug fix for [issue 1394](https://github.com/seek4science/seek/issues/1394):
- Adds creator field to the form when disigning an isa-study / isa-assay
- Adds creators to the `study` object when created from single page
- Adds creators to the `assay` object when created from single page

### DIFF
--- a/app/controllers/isa_assays_controller.rb
+++ b/app/controllers/isa_assays_controller.rb
@@ -14,7 +14,8 @@ class IsaAssaysController < ApplicationController
     update_sharing_policies @isa_assay.assay
     @isa_assay.assay.contributor = current_person
     @isa_assay.sample_type.contributor = User.current_user.person
-
+    @isa_assay.assay.creators = get_creator_ids_for_isa_assays(params).collect{ |creator_id| Person.find(creator_id) }
+    @isa_assay.assay.other_creators = params[:assay][:other_creators]
     if @isa_assay.save
       redirect_to single_page_path(id: @isa_assay.assay.projects.first, item_type: 'assay',
                                    item_id: @isa_assay.assay, notice: 'The ISA assay was created successfully!')
@@ -56,6 +57,15 @@ class IsaAssaysController < ApplicationController
   end
 
   private
+
+  def get_creator_ids_for_isa_assays(params = {})
+    creator_ids = []
+    creator_keys = params[:assay][:assets_creators_attributes].keys
+    creator_keys.each do |key|
+      creator_ids.append(params[:assay][:assets_creators_attributes][key][:creator_id])
+    end
+    creator_ids.map(&:to_i)
+end
 
   def isa_assay_params
     # TODO: get the params from a shared module
--- a/app/controllers/isa_assays_controller.rb
+++ b/app/controllers/isa_assays_controller.rb
@@ -15,7 +15,7 @@ class IsaAssaysController < ApplicationController
     @isa_assay.assay.contributor = current_person
     @isa_assay.sample_type.contributor = User.current_user.person
     @isa_assay.assay.creators = get_creator_ids_for_isa_assays(params).collect{ |creator_id| Person.find(creator_id) }
-    @isa_assay.assay.other_creators = params[:assay][:other_creators]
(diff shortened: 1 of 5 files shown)


app/controllers/isa_assays_controller.rb
```diff
@@ -14,7 +14,8 @@ def create
     update_sharing_policies @isa_assay.assay
     @isa_assay.assay.contributor = current_person
     @isa_assay.sample_type.contributor = User.current_user.person
-
+    @isa_assay.assay.creators = get_creator_ids_for_isa_assays(params).collect{ |creator_id| Person.find(creator_id) }
+    @isa_assay.assay.other_creators = params[:assay][:other_creators] if params[:assay].respond_to? :other_creators
```
stuzart, line 18: you shouldn't be needing to jump through these hoops if the param attributes are correct, which they should be if using the `assets/manage_specific_attributes` partial, which you seem to be. It should be being set when calling `IsaAssay.new(isa_assay_params)` or maybe calling update on `@isa_assay.assay`

I would double check they are getting passed through and approved in `isa_assay_params`

stuzart (changes requested):
I've added a few comments, but I think there is something else going on that shouldn't require these changes. 
The default creator should be being set with: https://github.com/seek4science/seek/blob/5f2fc53bfe87d4cd71386c98d043766c0cdb90b3/app/views/assets/_author_form.html.erb#L14

so it seems a problem with the attribute names, which you seem to be trying to work around. It suggests when created no creators are assigned regardless of if they are the submitter or not.

I notice in update there is

```
@isa_assay.assay.attributes = isa_assay_params[:assay]
```

and that may be all you needed to do in the create method too. Calling `@isa_assay.assay.update(isa_assay_params[:assay])` could be safer, as it would wipe out other attributes.

stuzart:
> ['new','provide_metadata']

perhaps you just need to add your single page action here

app/controllers/isa_assays_controller.rb
kdp-cloud: For a reason I cant't figure out the assay_params, are not fully passed to the isa_assay_params.

```ruby
@isa_assay.assay.attributes = isa_assay_params[:assay]
```
Sets the "other_creators" attribute but not the creators.

kdp-cloud: I've  found something peculiar: In the params there is a `params[:study]` and a `params[:isa_study][:study]` which contain different attributes.

```ruby
>> params[:study]
=> #<ActionController::Parameters {"investigation_id"=>"2", "assets_creators_attributes"=>{"1"=>{"pos"=>"1", "creator_id"=>"1", "family_name"=>"De Pelseneer", "given_name"=>"Kevin", "affiliation"=>"Default Institution, Fun University", "orcid"=>""}, "2"=>{"pos"=>"2", "creator_id"=>"3", "family_name"=>"Sheraton", "given_name"=>"Sheryl", "affiliation"=>"", "orcid"=>""}}, "other_creators"=>"Lord Bob Bobbington the third"} permitted: false>
>> params[:isa_study][:study]
=> #<ActionController::Parameters {"title"=>"Test study from Single Page", "investigation_id"=>"2", "description"=>"", "position"=>"", "publication_ids"=>[""], "sop_id"=>"1"} permitted: false>
```
I think that might be the reason the creators aren't linked "out of the box" and I have to manually assign them.

stuzart: I think I know why, in the `_author_form `it uses `form.object`  and bases the field names from that type, so will collapse it down to just `:study`

https://github.com/seek4science/seek/blob/cfb9dd0719bd86084d99219dd3b713109b302d8a/app/views/assets/_author_form.html.erb#L7

stuzart: I think this would be solvable by basing the `field_name `from `form.object_name` as instead, which will the parent, i.e

```
0> form.object_name
=> "isa_assay[assay]"
```

kdp-cloud: Thanks for the fix!
It all works out of the box now.

Issue #1470, opened 2023-05-31
Workflow with DOI was deleted


fbacall:
https://workflowhub.eu/workflows/379?version=1

Should not have been possible

PR #1475, opened 2023-06-02
Versioning fixes

### LINKED ISSUE #1229: Can't change version visibility for `Git::Version`s
It's possible in the model, but there's no UI or controller action to do it.

fbacall:
Some method renaming in versioning systems to prevent confusion.

- Methods prefixed with `standard_` always refer to `ExplicitVersioning` versions.
- Methods prefixed with `git_` always refer to git versions.
- Non-prefixed methods will switch between the 2 systems: https://github.com/seek4science/seek/blob/fix-versioning-ambiguity/lib/git/versioning_compatibility.rb
 
Fixes #1229 and #1470

Also fixes some types missing from `Person#contributed_items`.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -131,6 +131,8 @@ SEEK::Application.routes.draw do
         patch 'blob/*path' => 'git#move_file', as: :git_move_file
         get 'freeze' => 'git#freeze_preview', as: :git_freeze_preview
         post 'freeze' => 'git#freeze', as: :git_freeze
+        patch '' => 'git#update', as: :git_update_version
+        #delete '' => 'git#destroy', as: :git_destroy_version
       end
     end
   end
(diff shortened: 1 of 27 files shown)


config/routes.rb
```diff
@@ -131,6 +131,8 @@
         patch 'blob/*path' => 'git#move_file', as: :git_move_file
         get 'freeze' => 'git#freeze_preview', as: :git_freeze_preview
         post 'freeze' => 'git#freeze', as: :git_freeze
+        patch '' => 'git#update', as: :git_update_version
+        #delete '' => 'git#destroy', as: :git_destroy_version
```
stuzart, line 135: is this intended to be commented out, and if so could just be removed ?

stuzart (changes requested):
just something commented out to double check

config/routes.rb
fbacall: Yeah it can probably be removed. I started looking into how removing versions works for non-git versioned things but wasn't sure if we are supporting that feature now that you can just hide versions.

stuzart (approved):

stuzart:
Although I approved it, I've just noticed the unit tests are failing.

PR #1497, opened 2023-06-27
Use CFF to generate citation if present

### LINKED ISSUE #1160: Amend citation
There are a few issues with how citations appear in workflowhub. I'd like to suggest using Github's CITATION.cff file to populate the citation box in WorkflowHub. This would resolve these issues including: 

- Authors do not appear in the order specified in Manage Workflow 
- Not all authors appear, despite being listed as creators. 
- We would like to include authors who do not have a workflowhub account
- It is missing a few fields required by some citation styles, including version of the workflow and date released

Using CITATION.cff would also ensure consistency between the two platforms. 

Here's an example: https://workflowhub.eu/workflows/146

fbacall:
If workflow has a CITATION.cff file but no DOI, use the metadata from CFF file to display the citation.

Fixes #1160

### DIFF
--- a/app/assets/javascripts/application_shared.js.erb
+++ b/app/assets/javascripts/application_shared.js.erb
@@ -247,9 +247,9 @@ $j(document).ready(function () {
         var style = $j(this).val();
         var url = $j(this).data('url');
         var target = $j('#citation');
-        var currentStyle = $j('#citation').data('citationStyle');
+        var currentStyle = $j('#citation div').data('citationStyle');
 
-        if (style && style !== currentStyle) {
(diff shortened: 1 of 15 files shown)


db/schema.rb
stuzart, line 1: Looks like the schema needs reverting, there aren't any migrations but lots of extra junk has been added

stuzart (changes requested):

db/schema.rb
fbacall: Yeah, didn't mean to commit that.

stuzart (approved):

PR #1501, opened 2023-07-04
License extraction

fbacall:
Detect and parse LICENSE etc. file in a workflow's git repo/RO-Crate and return a matching SPDX ID.

Also DRYs up some git repo/RO-Crate extraction code.


Gemfile.lock
```diff
@@ -927,7 +939,7 @@ GEM
       rubyzip (~> 2.0.0)
 
 PLATFORMS
-  ruby
```
stuzart, line 930: Platform has changed and needs locking for Ruby

stuzart (changes requested):

Gemfile.lock
fbacall: is there a way to do this without hacking Gemfile.lock?

stuzart: I had to do
```
 bundle lock --add-platform ruby
 bundle lock --remove-platform x86_64-linux
```
the remove-platform might not be necesary for you

PR #1502, opened 2023-07-05
Automatically expire footer cache when settings changed

fbacall:
Uses Rails' cache key generation to detect if settings have changed

Also update Manchester/HITS website links to HTTPS

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -1,58 +1,55 @@
-<% logged_in?? cache_name = 'footer' : cache_name = 'footer1' %>
-
-
-    <div id="ft">
-      <div class="container">
-        <div id="seek-credit">
-          <a href="http://www.seek4science.org" target="_blank" class="powered-by">
-            Powered by<br/>
-            <%= image(:seek_logo, width: 64, height: 64) %>
-          </a>
-          <span class="subtle"><%= version_text -%></span>
-        </div>
-        <% cache cache_name, skip_digest: true do -%>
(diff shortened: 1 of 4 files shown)


app/views/layouts/_footer.html.erb
```diff
@@ -1,58 +1,55 @@
...
+        <% if (Seek::Config.instance_admins_link != Seek::Config.instance_link) && (Seek::Config.instance_admins_name != Seek::Config.instance_name) %>
+          <%= link_to "About #{Seek::Config.instance_admins_name}", Seek::Config.instance_admins_link, target: '_blank' %> |
+        <% end %>
+        <%= link_to "About #{Seek::Config.instance_name}", Seek::Config.instance_link, target: '_blank' %> |
+        <% if logged_in_and_registered? %>
+          <%= link_to 'Provide feedback', feedback_home_path -%> |
+        <% end %>
+        <%= link_to 'Funding and Programmes', funding_home_path -%> |
+        <% if Seek::Config.funding_link.present? %>
+          <a href="<%=Seek::Config.funding_link%>" target="_blank">
```
stuzart, line 21: this should be changed to `link_to` (no idea why it was written this way). I don't think link_to automatically adds the noreferrer though. We may need to update all links that open in a new tab.

stuzart (changes requested):

app/views/layouts/_footer.html.erb
fbacall: i'll just change it to `link_to` for now, we can think about how to update all external links later

PR #1512, opened 2023-07-12
Group child annotations and display count attempt 2

### LINKED ISSUE #1461: Git directory with multiple remote files inside shows duplicate annotations
image

only needs to show 1 of each type.

fbacall:
Fixes #1461 

Top level
image

`test` expanded
image

`test/123` expanded
image

### DIFF
--- a/app/assets/javascripts/git.js.erb
+++ b/app/assets/javascripts/git.js.erb
@@ -32,10 +32,20 @@ $j(document).ready(function () {
 
                     if (tree.is_closed(node) && node.data.childAnnotations && node.data.childAnnotations.length) {
                         element.addClass('git-annotated-parent');
+                        var groups = {};
                         node.data.childAnnotations.forEach(function(annotation) {
-                            var template = HandlebarsTemplates["git/annotation"];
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/git.js.erb
```diff
@@ -32,10 +32,24 @@ $j(document).ready(function () {
...
+                                }
+                                groups[annotation.key].count++;
+                            }
                         });
+                        for (var key in groups) {
+                            if (groups.hasOwnProperty(key)) {
+                                var template = HandlebarsTemplates["git/annotation"];
+                                element.append(template(groups[key]));
+                            }
+                        }
```
stuzart, line 52: pure pedantry, and only because I've just been doing this with the Nels javascript - but some of these `var` definitions could be `const` or `let`
Not a blocker though

stuzart (approved):

app/assets/javascripts/git.js.erb
fbacall: Yeah I just kept it consistent with the other code. We should do an ES6 refactor at some point.

PR #1520, opened 2023-07-13
Upload samples through spreadsheet single page

kdp-cloud:
Contains functionality to create and update samples directly by uploading an Excel Spreadsheet.


test/functional/single_pages_controller_test.rb
```diff
@@ -181,11 +98,238 @@ def setup
...
+  end
+
+  test 'Should update, create and detect duplicate sources when uploading to a source Sample Type' do
+    with_config_value(:project_single_page_enabled, true) do
+      project, source_sample_type = setup_file_upload.values_at(
+        :project, :source_sample_type
+      )
+
+      file_path = Rails.root.join('test/fixtures/files/upload_single_page/01_combo_update_sources_spreadsheet.xlsx')
+      file = fixture_file_upload(file_path, 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet')
```
stuzart, line 165: for the file_path you just need to pass the path relative to the `test/fixtures/files`, and it will do the join for you. So you just need to pass `upload_single_page/01_combo_update_sources_spreadsheet.xlsx`

app/controllers/single_pages_controller.rb
```diff
@@ -122,8 +123,201 @@ def export_to_excel
     end
   end
 
+  def upload_samples
```
stuzart, line 126: this is a hell of a long method that would benefit and be easier to understand if split up, or made use of some classes

app/views/single_pages/sample_upload_content.html.erb
```diff
@@ -0,0 +1,405 @@
...
+  def any_samples_to_upload?
+    !((@new_samples.nil? or @new_samples.compact.none?) and (@update_samples.nil? or @update_samples.compact.none?) and (@possible_duplicates.nil? or @possible_duplicates.compact.none?))
+  end
+  errors = []
+%>
+
+<% can_upload = true %>
+<div id="upload-summary">
+
+  <% # General information panel %>
```
stuzart, line 11: ERB comments should start with `<%#` i.e. no space after the %

app/views/single_pages/sample_upload_content.html.erb
```diff
@@ -0,0 +1,405 @@
+<%
```
stuzart, line 1: there is a lot of code directly in the single view here, and would be easier to understand if split into partials and/or view helper methods

stuzart (approved):
A few comments, but no blockers

Issue #1452, opened 2023-05-16
Show workflows of group shows top 5


jjkoehorst:
**Describe the bug**
When going to the workflow page of a group it only shows the top 5 and not the whole list.

**To Reproduce**
1. Go to https://workflowhub.eu/search?utf8=✓&q=unlock#workflows
2. Scroll down and count

**Expected behavior**
A next page or a full list (which might be a problem if you have 100s of workflows...

**Screenshots**

https://workflowhub.eu/search?utf8=✓&q=unlock#workflows

stuzart:
there is a link to the full list along with additional filtering at the top, but I think this needs to be clearer. 
https://workflowhub.eu/workflows?filter%5Bquery%5D=unlock

stuzart:
there is also a configuration that can set the default list length to be greater than 5, I've just updated it to 20 for the workflowhub which matches fairdomhub

jjkoehorst:
Do you mean the [Advanced Workflows search with filtering ...](https://workflowhub.eu/workflows?filter%5Bquery%5D=unlock) button? as indeed it shows everything the link itself does not state that...

stuzart:
I will start addressing this with 2 changes, if there more results than shown it will be changed to

- where and when it says, e.g. "Showing 20 out of a possible 180 items", the 180 items will be changed to a link to the full results
- if there are more results than displayed, then a link will be shown at the bottom that takes you to the full results, the same as Advanced Workflows search with filtering but with different wording, such as "All results with advanced filtering"

PR #1527, opened 2023-07-17
Make it clearer there are additional search results

stuzart:
Made it clearer there are more search results than shown, and how get to them. Also applies to related items. As described on the issue #1452 :

- where and when it says, e.g. "Showing 20 out of a possible 180 items", the 180 items will be changed to a link to the full results
- if there are more results than displayed, then a link will be shown at the bottom that takes you to the full results, the same as Advanced Workflows search with filtering but with different wording, such as "All results with advanced filtering"

### DIFF
--- a/app/helpers/resource_list_helper.rb
+++ b/app/helpers/resource_list_helper.rb
@@ -59,4 +59,47 @@ module ResourceListHelper
       end
     end
   end
+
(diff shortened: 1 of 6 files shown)


fbacall (approved):
1 minor comment

Issue #456, opened 2020-11-03
RO-Crate "license" field should be URI


stain:
As pointed out in https://github.com/seek4science/seek/issues/183#issuecomment-721016513 the Workflow RO-Crate `license` field is in SEEK treated like a text field (e.g. `"MIT"`), however both https://www.researchobject.org/ro-crate/1.1/contextual-entities.html#licensing-access-control-and-copyright and https://schema.org/license says it should be a URL, meaning {"@id": "https://spdx.org/licenses/MIT"} or similar.

The Workflow RO-Crate `license` field should be valid and ideally also consistent with the full URIs in the generated schema.org annotations.

stain:
The below tries to capture a Skype conversation on 2020-11-03 between @stain @fbacall @stuzart @alaninmcr 

This would improve consistency with the schema.org generation, which currently use https://github.com/seek4science/seek/blob/master/lib/seek/license.rb to parse https://github.com/seek4science/seek/blob/master/public/od_licenses.json from https://licenses.opendefinition.org/licenses/groups/all.json aka https://opendefinition.org/licenses/api/

Although opendefinition.org and opensource.org use SPDX identifiers from https://spdx.org/ they have additional information such as if it's suitable for data or software - which is important in SEEK.

[SPDX](https://spdx.dev/resources/use/) is in Open Source considered the gold standard for license information, particularly the short SPDX identifiers embedded [inside source code files](https://spdx.dev/ids/#how) as comments (thus allowing different files to have different licenses):

```
# SPDX-License-Identifier: GPL-2.0-or-later
```

For packages or repositories SPDX also have something called [SPDX documents](https://spdx.github.io/spdx-spec/3-package-information/) which either have a simple yet unwieldy long text file, or an RDF document. In here we find [spdx:licenseId](https://spdx.org/rdf/ontology/spdx-2-2/#http://spdx.org/rdf/terms#licenseId) which we could in theory use from RO-Crate as a way to equalize across `@id`s. 

A potential gotcha here is that this would also allow [license expressions](https://spdx.dev/spdx-specification-21-web-version/#h.jxpfx0ykyb60) e.g. for dual licensing or exceptions - so these don't always match straight up into URLs https://spdx.org/licenses/MIT.

Many of the licenses have their own URIs as well, so we could have many potential inconsistencies:
* <https://opensource.org/licenses/Apache-2.0>
* <https://spdx.org/licenses/Apache-2.0> (as used in SPDX RDF)
[...]

PR #1546, opened 2023-07-25
SPDX support + license selection improvements

### LINKED ISSUE #114: Feature request: Licenses
Proposal: add more optional opensource licenses, e.g. MIT, GPLv3, ...listed in [https://opensource.org/licenses](https://opensource.org/licenses) -- browser drop-down and API supported; also enabled for Snapshots

### LINKED ISSUE #655: spdx and specify copyright holder
The license references should use a common mechanism, preferably the spdx site

fbacall:
- Adds SPDX licenses to the current supported set of licenses in SEEK (non-SPDX licenses are retained).
- Improves license selection to be searchable/filterable (using select2).
- Allows licenses to be provided by URI instead of just ID.

image

image

- Fixes #114
- Fixes #655
- Helps support #456


lib/seek/license.rb
```diff
@@ -2,55 +2,84 @@ module Seek
...
-      json
-    end
-
-    private_class_method def self.organize_licenses(licenses)
-      category = {}
+    private_class_method def self.parse_od(licenses)
+                           # overrides values taken from the JSON.
+                           # Preferable to modifying the JSON file directly which is a definitive source and may be replaced with an updated copy
+                           licenses['notspecified']['title'] = I18n.t('null_license')
+                           licenses['notspecified']['url'] = Seek::Help::HelpDictionary.instance.help_link(:null_license)
```
stuzart, line 9: should be using `NULL_LICENSE ` constant to get the key

app/helpers/license_helper.rb
```diff
@@ -3,15 +3,20 @@
...
+    opts[:data] ||= {}
+    opts[:data]['role'] ||= 'seek-license-select'
+    recommended = opts.delete(:recommended)
+    source = opts.delete(:source) || Seek::License.combined
+    if recommended
+      options = grouped_options_for_select(grouped_license_options(source.values, recommended), selected)
+    else
+      options = options_for_select(license_options(source.values), selected)
+    end
+    select_tag(name, options, opts)
```
stuzart, line 15: would be good if this could be done via the `objects_input` helper, in the interest of keeping things self contained and easier to understand and maintain in the future. I can see the grouped options would be a problem here, but may be something useful elsewhere in the future. Did you give this any thought?

I'm a bit uncomfortable having different `select2 `implementations in various places, it was a pain and error prone when switching over from `typeahead`

stuzart (changes requested):

app/helpers/license_helper.rb
fbacall: Could use `objects_input`, but I think it needs some tweaks to make it more appropriate for selecting single values.

app/assets/javascripts/licenses.js
```diff
@@ -0,0 +1,24 @@
...
+
+            if (element.data('url')) {
+                block.show();
+            } else {
+                block.hide();
+            }
+
+            link.attr('href', element.data('url'));
+            link.html(element.data('url'));
+        }
```
stuzart, line 22: might want to look at https://select2.org/programmatic-control/retrieving-selections about retrieving selections, which also advices against reading the selected option. Probably also use `select2:select` as the event for when it has been selected.

stuzart (changes requested):

app/helpers/license_helper.rb
fbacall: It's currently just a regular select tag (not a multiple select combobox) so it only allows 1 selection by default.

Limiting to 1 is a possibility, but then you have to clear the value with the X icon before you can choose a different license, instead of just directly selecting another option, which is a bit awkward

stuzart (approved):

PR #1561, opened 2023-08-09
Issue 1535 Fix sharing permission of Assay samples

kdp-cloud:
This PR contains a fix for issue 1535.

When selecting the samples in an assay table in single page, it selected the samples in the source samples table instead.

### DIFF
--- a/app/assets/javascripts/single_page/dynamic_table.js.erb
+++ b/app/assets/javascripts/single_page/dynamic_table.js.erb
@@ -35,7 +35,7 @@ const handleSelect = (e) => {
     dtSelectAll.html("Select all rows")
     dtSelectAll.css("display", "none")
   }
-  const dtName = $j(e).parents("table").data("dtname")
+  const dtName = $j(e).parents("table").data("dtname");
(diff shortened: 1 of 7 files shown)


fbacall (commented):
Which change here actually fixed the issue?

The only functional change I can see is `var` instead of `let`, but I'm not sure how that would change the behaviour.

kdp-cloud:
I agree it's trivial but it solved the problem. Here's my shot at a theory:
Using `let`  limits the scope of the variable `sampleDynamicTable`, making it unavailable when `setPermissionBTNActivity` is called in [dynamic_table.js.erb](https://github.com/seek4science/seek/blob/d9cee8c48a43700c013e355fa6bde1db579e18eb/app/assets/javascripts/single_page/dynamic_table.js.erb#L39) . When you use `var` instead, it stays into scope and allows `setPermissionBTNActivity` to be called on the variable.

fbacall:
In that case I think it would be good to explicitly assign the variable to `window`, e.g. `window.sampleDynamicTable = ...`, and always reference it that way.

fbacall (approved):

PR #1564, opened 2023-08-15
Seek 1.14 multi custom metadata

whomingbird:
1. added support for "Linked Custom Metadata (Multiple)" Custom Metadata Attribute Type: which enables a custom metadata attribute to link to an array of custom metadata entries.

2. restructured code for the features of Linked Custom Metadata and Linked Custom Metadata (
multiple) Custom Metadata Attribute Type : a new approach has been implemented to store nested custom metadata
within a single record ( Many thanks to Finn, he helped me to rewrite the controller and validator)


app/helpers/samples_helper.rb
```diff
@@ -41,29 +41,25 @@ def controlled_vocab_list_form_field(sample_controlled_vocab, element_name, valu
...
   end
 
-  def linked_custom_metadata_form_field(attribute,resource,element_name, element_class,depth)
-    linked_cms = resource.linked_custom_metadatas.select{|cm|cm.custom_metadata_attribute==attribute}
-
-    id = linked_cms.blank? ? nil : linked_cms.select{|cm| cm.custom_metadata_type.id == attribute.linked_custom_metadata_type.id}.first.id
+  def linked_custom_metadata_multi_form_field(attribute, value, element_name, element_class)
+    render :partial => 'custom_metadata/fancy_linked_custom_metadata_multi_attribute_fields',
+           :locals=> {:value=>value, :attribute => attribute,:element_name => element_name,:element_class => element_class, :collapsed=>false}
+  end
```
stuzart, line 47: use `partial:` as the keys, rather than the older fashioned `:partial=>` (applies to the other keys too, not just partial)

app/views/custom_metadata/_custom_metadata_attribute_values.html.erb
```diff
@@ -8,15 +8,22 @@
...
+                <%= folding_panel(attribute.label, true, id:attribute.title) do %>
+                  <%= display_attribute(resource.custom_metadata, attribute, link: true) %>
+                <% end %>
+              <% end %>
+            <% else %>
+              <label><%= attribute.label %>:</label>
+              <%= display_attribute(resource.custom_metadata, attribute, link: true) %>
+            <% end %>
+          </div>
+        <% end %>
```
stuzart, line 24: this block of code would be better and easier to understand in a helper method

app/views/custom_metadata/_fancy_linked_custom_metadata_multi_attribute_fields.html.erb
```diff
@@ -0,0 +1,62 @@
...
+            });
+            index++;
+            newRow = newRow.replace(/row-template/g, index);
+
+            if (lastRow.length == 0){
+                container.prepend(newRow);
+            }else{
+                // find the last row of this kind and the new row after that
+                container.find('#<%= prefix %>-role-'+(index-1)).after($j(newRow));
+            }
```
stuzart, line 57: this would be a bit cleaner as a function that took `key.singularize` and `prefix` as a parameter, to avoid multiple cases of `<%= key.singularize` %>`. Or you could just define them as variables at the start of the block of code

app/models/sample_attribute.rb
```diff
@@ -53,6 +53,10 @@ def short_pid
     URI.parse(pid).fragment || pid.gsub(/.*\//,'') || pid
   end
 
+  def linked_custom_metadata_type
+    nil
+  end
+
```
stuzart, line 59: I'm not sure why this is needed? Looks a bit fishy

stuzart (changes requested):
I think it would benefit from more unit tests around the models. It is quite complex and would be easier to understand and debug in the future via unit tests rather than functional. Particulary around validation and how the data is stored

app/models/sample_attribute.rb
fbacall: It's the equivalent of https://github.com/seek4science/seek/blob/9ef544c92b9334156287c353cd0ef9cd1f475c96/app/models/custom_metadata_attribute.rb#L13-L15

app/models/custom_metadata_type.rb
```diff
@@ -23,6 +23,7 @@ def attributes_with_linked_custom_metadata_type
 
   def supported_type_must_be_valid_type
     return if supported_type.blank? # already convered by presence validation
+    return if supported_type == "CustomMetadata"
```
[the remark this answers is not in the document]
whomingbird: Can I remove this line but add 
`  Seek::Config.default :custom_metadata_enabled, true`
which will make 
`Seek::Util.lookup_class(supported_type, raise: false)` return true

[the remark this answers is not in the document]
whomingbird: also need to add `custom_metadata` in the `module EnabledFeaturesFilter`

lib/seek/enabled_features_filter.rb
```diff
@@ -1,6 +1,6 @@
 module Seek
   module EnabledFeaturesFilter
-    FEATURES = %i[assays biosamples documentation events models
+    FEATURES = %i[assays biosamples documentation events models custom_metadata
```
stuzart, line 3: this will cause pages to break if custom_metadata is disabled

lib/seek/config_setting_attributes.yml
```diff
@@ -231,7 +231,7 @@ omniauth_github_client_id:
 omniauth_github_secret:
   encrypt: true
 ga4gh_trs_api_enabled:
-custom_metadata_enabled: false
+custom_metadata_enabled:
```
stuzart, line 234: we don't want this as a configuration, unless the views are updated to honour it, otherwise things will break
The previous line is an old artifact where it was added and then removed, choosing instead to be determined by whether any custom metadata is configured, and in the future be able to turn of individual custom metadata definitions

stuzart (changes requested):

lib/seek/config_setting_attributes.yml
whomingbird: The reason for making this change is that I removed the line  `return if supported_type == "CustomMetadata"`
from [app/models/custom_metadata_type.rb](https://github.com/seek4science/seek/pull/1564/files/9ef544c92b9334156287c353cd0ef9cd1f475c96#diff-5597508804483c32920f9c0f88bad920d703e4bf6c35f8a9d65db7da9e547b95)

I introduced the new supported type, "CustomMetadata," to distinguish between custom metadata directly related to ISA items and linked custom metadata. Removing this line resulted in the failure of CustomMetadataType creation when a linkedCustomMetadata type was present.

app/models/custom_metadata_type.rb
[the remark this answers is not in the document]
stuzart: I think your original check was preferable to turning it into a configuration option
or you could override `self.feature_enabled?` for `CustomMetadata`

stuzart (approved):

PR #1567, opened 2023-08-18
Sample and Multi Sample form element

### LINKED ISSUE #1557: Difference between "Registered sample" attribute type and its variation "multiple"
Attribute types "Registered sample" and its "multiple" variation behave differently.

Use case: User selects the sample "Vanilla test Sample 3" as value for a registered sample attribute for a sample. If the referenced sample "Vanilla test Sample 3" gets Hidden or Deleted, the referenced sample is
- in "Registered sample": not visible nor selectable from the dropdown menu for the User
- in "Registered sample multiple": STILL visible AND selectable from the dropdown menu for the User. This seems a bug; this should behave as "Registered sample".

Screenshot 2023-08-09 at 11 43 38

stuzart:
Use select2 for both Single and Multiple Sample entry (requires handling params as an array for both cases, whereas the api still takes a non array for single case)

Also make sure non visible samples are filtered from the available options.

Also checked that non visible samples are shown as hidden when displaying the parent sample

- Fixes #1557

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -614,7 +614,7 @@ class ApplicationController < ActionController::Base
         keys = [:custom_metadata_type_id,:id,:custom_metadata_attribute_id]
         cma= []
         metadata_type.custom_metadata_attributes.each do |attr|
-          if attr.sample_attribute_type.controlled_vocab? || attr.sample_attribute_type.seek_sample_multi?
+          if attr.sample_attribute_type.controlled_vocab? || attr.sample_attribute_type.seek_sample_multi? || attr.sample_attribute_type.seek_sample?
             cma << {attr.title=>[]}
             cma << attr.title.to_s
           elsif  attr.linked_custom_metadata?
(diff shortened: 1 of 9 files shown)


fbacall (commented):
What happens if a user creates a sample, and links a private sample that only they can view, then another user (with edit permissions) edits the original sample - can they see the private sample? 

and if not, does it unlink the private sample when they submit the form?

floradanna:
To the best of my knowledge, the hidden linked sample remains labelled as "hidden" but the other user can still edit the sample if he/she has editing rights over the sample.

stuzart:
> What happens if a user creates a sample, and links a private sample that only they can view, then another user (with edit permissions) edits the original sample - can they see the private sample?
> 
> and if not, does it unlink the private sample when they submit the form?

I tried testing it, and the private sample is inculded and it's title revealed. It isn't stripped out when resubmitting. I'm changing it so that the title is kept hidden.

floradanna:
Indeed, **in the form** the title is revealed for registered samples (multiple), but not for registered sample (single). See #1557 
When looking at the sample instead, the label "hidden" is shown.

fbacall (approved):

PR #1571, opened 2023-08-22
Issue 1457 delete sample types with deleted assets

kdp-cloud:
This Pull Request fixes:
- Deletes the linked sample types of the deleted assays or studies
- Takes into account the samples in the linked sample types when checking at deletion.

### DIFF
--- a/app/controllers/assays_controller.rb
+++ b/app/controllers/assays_controller.rb
@@ -18,7 +18,7 @@ class AssaysController < ApplicationController
   api_actions :index, :show, :create, :update, :destroy
 
   def new_object_based_on_existing_one
-    @existing_assay =  Assay.find(params[:id])
+    @existing_assay = Assay.find(params[:id])
     @assay = @existing_assay.clone_with_associations
 
     if @existing_assay.can_view?
@@ -62,9 +62,7 @@ class AssaysController < ApplicationController
       flash[:error]="You do not have the necessary permissions to copy this #{t('assays.assay')}"
       redirect_to @existing_assay
     end
-
-
-   end
+  end
 
   def new
     @assay=setup_new_asset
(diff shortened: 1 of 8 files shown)


app/controllers/assays_controller.rb
```diff
@@ -112,6 +111,13 @@ def create
     end
   end
 
+
+  def delete_linked_sample_types
+    return unless is_single_page_assay?
+
+    @assay.sample_type.delete
```
stuzart, line 118: you should use `.destroy` to ensure any callbacks are called. `delete `will just delete the record from the database without any checks

stuzart (changes requested):
I think it also needs some unit tests, to check for the model states of whether they can be deleted

app/controllers/assays_controller.rb
kdp-cloud: Thanks, didn't know that. See commit [eed9e1c](https://github.com/seek4science/seek/pull/1571/commits/eed9e1c917c301a49193a01bb1cea3fd3bed4ad3).

app/controllers/studies_controller.rb
```diff
@@ -88,6 +89,17 @@ def update
     end
   end
 
+  def delete_linked_sample_types
+    return unless is_single_page_study?
+
+    @study.sample_types.each do |st|
+      raise "Sample Type '#{st.title}' contains samples. Unable to delete study" unless st.samples.empty?
+
+      st.delete
```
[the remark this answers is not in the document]
kdp-cloud: See commit [eed9e1c](https://github.com/seek4science/seek/pull/1571/commits/eed9e1c917c301a49193a01bb1cea3fd3bed4ad3).

stuzart (approved):

PR #1578, opened 2023-08-24
Sitemaps

### LINKED ISSUE #1250: Generate Sitemap for list of public content
**Describe the bug**
Programmatic clients such as search engines and FAIR clients have no way to iterate all the content of WorkflowHub or to see when some of the assets may have been updated since last visit.

**Additional context**

[Sitemaps](https://www.sitemaps.org/protocol.html) is a method promoted by search engines that simply list pages and their last modification date. It is possible to have nested sitemaps, e.g. TeSS does it like this https://tess.elixir-europe.org/sitemap.xml per asset type + other pages.

**Other alternatives**

* Each asset collection page e.g. https://workflowhub.eu/workflows could use a schema.org type like https://workflowhub.eu/workflows -- unsure how they will be listed under https://workflowhub.eu/
* [OAI-PMH](https://www.openarchives.org/pmh/) is a classic used by harvesters like OpenAIRE and exposed by repositories like Zenodo
* [SWORD v3](https://sword.cottagelabs.com/swordv3/) is a modern alternative to the above that supports metadata and write operations. 
* .. I'm sure there are many more

fherreazcue:
Adds sitemaps using sitemap_generator gem.

Sitemaps are grouped by type, and only publicly visible items are added.

Resolves #1250

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -1,26 +1,19 @@
-# See the readme at https://github.com/lassebunk/dynamic_sitemaps
-base_url = URI.parse(Seek::Config.site_base_url)
-host_with_port = base_url.host
-if base_url.port != base_url.default_port
-  host_with_port += ":#{base_url.port}"
-end
+# https://github.com/kjvarga/sitemap_generator#sitemapgenerator
+SitemapGenerator::Sitemap.sitemaps_path = "sitemaps"
+SitemapGenerator::Sitemap.create_index = "auto"
+SitemapGenerator::Sitemap.compress = false
+SitemapGenerator::Sitemap.default_host = URI.parse(Seek::Config.site_base_url)
 
-protocol base_url.scheme
-host host_with_port
-
-# You can have multiple sitemaps – just make sure their names are different.
-sitemap :site do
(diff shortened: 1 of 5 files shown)


config/sitemap.rb
```diff
@@ -0,0 +1,19 @@
...
+SitemapGenerator::Sitemap.create do
+  Seek::Util.searchable_types.each do |type|
+    add  polymorphic_path(type), lastmod: type.maximum(:updated_at), changefreq: 'daily', priority: 0.7
+  end
+end
+
+Seek::Util.searchable_types.each do |type|
+  SitemapGenerator::Sitemap.create(filename: type.table_name, include_root: false) do
+    type.authorized_for('view', nil).find_all do |obj|
+      add polymorphic_path(obj), lastmod: type.maximum(:updated_at), changefreq: 'daily', priority: 0.7
```
fbacall, line 16: `lastmod` for the object should be its own `updated_at`: `obj.updated_at`

fherreazcue: Fixed in [4d68328](https://github.com/seek4science/seek/pull/1578/commits/4d68328bdbcde9a2cdf250893d9b498f8c70d9b2).

stuzart (approved):

Issue #1357, opened 2023-02-14
Make samples extracted from datafile editable


floradanna:
Problem: samples extracted from a datafile cannot be edited because they are forever bound to the information provided in the datafile.

Solution: samples extracted from a datafile should be editable after extraction, as any other samples in SEEK.

Process: 
- the user goes to "edit sample",
- if sample has been extracted from a datafile, SEEK shows a warning "Sample has been extracted from a datafile, editing this sample will make the datafile not valid anymore for this sample" or something.
- if user continue editing, the datafile link for that sample will gain the label "not valid anymore"

Process for editing a **sample type** that has a linked datafile: same process as described for the editing of samples

stuzart:
this relates to #1136 and involves decoupling the permissions and projects from the datafile. The datafile could just provide the defaults

stuzart:
also #1356

PR #1599, opened 2023-09-29
Editable extracted samples

### LINKED ISSUE #1356: Decouple sharing permissions of datafile from extracted samples
Problem: samples extracted from datafile are bound to the sharing permission settings of the datafile.

Solution: decouple sharing permissions of datafile from the sharing permission of the extracted samples

Process: while extracting samples from datafile, the user should define new sharing permission for the extracted samples. Possible sharing permissions to be applied could be:
- default permissions, if a default policy for the entire project is set
- other --> batch change of sharing permission for samples

Add metadata field for sample ("original datafile" or similar) to keep the link between the sample and the datafile used for extracting it.

Related to #1136

fherreazcue:
Makes extracted samples editable and decouples their policy from that of the source datafile.
Resolves #1356 and #1357 .

### DIFF
--- a/test/unit/policy_test.rb
+++ b/test/unit/policy_test.rb
@@ -194,37 +194,39 @@ class PolicyTest < ActiveSupport::TestCase
     end
   end
 
-  test 'policy not destroyed if still referenced by assets' do
-    policy = FactoryBot.create(:public_policy)
-    sample_type = FactoryBot.create(:strain_sample_type)
-    data_file = FactoryBot.create(:strain_sample_data_file, policy: policy)
-    samples = data_file.extract_samples(sample_type, true).select(&:persisted?)
-    sample = samples.first
-
-    assert_equal sample.policy, data_file.policy
-
-    assert_no_difference('Policy.count') do
-      disable_authorization_checks { data_file.destroy }
-    end
-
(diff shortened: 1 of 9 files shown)


test/unit/policy_test.rb
```diff
@@ -194,37 +194,39 @@ class PolicyTest < ActiveSupport::TestCase
...
-
-    disable_authorization_checks { data_file.destroy }
-
-    assert_difference('Policy.count', -1) do
-      disable_authorization_checks { samples.each(&:destroy) }
-    end
-
-    assert_nil Policy.find_by_id(policy.id)
-  end
+  # Samples no longer reference the policy, they make a copy of it.
```
stuzart, line 197: delete these tests, rather than leave commented out

stuzart (changes requested):

stuzart (approved):

test/unit/policy_test.rb
fherreazcue: Instead of deleting them I replaced samples with two sops forced to reference same policy in [7bac631](https://github.com/seek4science/seek/pull/1599/commits/7bac631a6b0394eff078f81fc3e1c13a1611a289).

Issue #1602, opened 2023-10-04
DataHub Configuration - add Template to SP advanced


floradanna:
The feature in the admin panel "Sample type Template enabled" (Whether users can create and manage Templates for Sample types) should be part of "Advanced features enabled" (If set, the advance features of the single page is enabled.)

Instance admin should not be able to enable Template without SinglePage advanced.

somathias:
Should instance admins be able to enable SinglePage advanced without Templates?

floradanna:
Instance admins should NOT be able to enable SinglePage advanced without Templates.

floradanna:
related to #1604

PR #1605, opened 2023-10-05
Add a show_query_button flag to the general index view

### LINKED ISSUE #1604: "Query for Template" button visible for all assets, not only Samples
**Describe the bug**
When browsing any asset or ISA element, the "Query for Template" button that leads to the querying of Samples by Templates is displayed next to the new button.

**To Reproduce**
1. Go to 'Browse' > Investigations
2. See button

**Expected behavior**
The button should only appear on the Samples index page.

**Screenshots**
image

somathias:
Closes #1604

### DIFF
--- a/app/views/general/_index.html.erb
+++ b/app/views/general/_index.html.erb
@@ -3,16 +3,18 @@
   new_item_path = polymorphic_path([:new, controller_name.singularize.to_sym])
   items ||= instance_variable_get("@#{controller_name}")
   show_new_button = true unless local_assigns.has_key?(:show_new_button)
+  show_query_button = false unless local_assigns.has_key?(:show_query_button)
   title ||= nil
   subtitle ||= nil
 	isa_templates_enabled = Seek::Config.sample_type_template_enabled && Seek::Config.project_single_page_advanced_enabled
 %>
 
 <div class="pull-right">
+  <% if show_query_button && isa_templates_enabled %>
+    <%= link_to "Query by #{t('template').pluralize}", query_form_samples_path, class: "btn btn-default btn" %>
+  <% end %>
+
   <% if show_new_button && controller_model.can_create? %>
-    <% if isa_templates_enabled %>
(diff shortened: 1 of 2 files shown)


kdp-cloud (approved):

app/views/general/_index.html.erb
```diff
@@ -3,16 +3,18 @@
...
   items ||= instance_variable_get("@#{controller_name}")
   show_new_button = true unless local_assigns.has_key?(:show_new_button)
+  show_query_button = false unless local_assigns.has_key?(:show_query_button)
   title ||= nil
   subtitle ||= nil
 	isa_templates_enabled = Seek::Config.sample_type_template_enabled && Seek::Config.project_single_page_advanced_enabled
 %>
 
 <div class="pull-right">
+  <% if show_query_button && isa_templates_enabled %>
```
stuzart, line 13: it would clearer to use `isa_templates_enabled` to pass and set `show_query_button` flag
the flag should also have a clearer name, since this page is used for all types - e.g `show_samples_query_button`

app/views/samples/index.html.erb
```diff
@@ -25,5 +25,5 @@
       </script>
   <% end %>
 <% else %>
-    <%= render partial: "general/index" %>
+    <%= render partial: "general/index",:locals=>{:show_query_button=>true} %>
```
stuzart, line 28: as mentioned, instead of always passing true, it should pass `isa_templates_enabled`

stuzart (changes requested):

app/views/general/_index.html.erb
somathias: Agreed. Thanks for the comments.
This change request is in line with the plans to collect all ISA-JSON export-related features (including ST Templates and what is now called SP advanced) under a single configuration option, see discussions on issues #1602 and #1604 .

somathias:
The requested changes will be covered in PR #1621

PR #1606, opened 2023-10-05
rename custom metadata to extended metadata....  

whomingbird:
1. update model definitions 
2. new migrations to rename the column.
3. update Views and Controllers.
4. modify tests.
5. update routes


app/models/concerns/has_extended_metadata.rb
```diff
@@ -0,0 +1,28 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    def extended_metadata_attribute_values_for_search
+      extended_metadata ? extended_metadata.data.values.reject(&:blank?).uniq : []
+    end
+  end
+
+  class_methods do
+    def has_extended_extended_metadata
```
stuzart, line 11: this can just be `has_extended_metadata`

stuzart (changes requested):
looks good, just one pedantic very minor change suggestion

will need a thorough test on the test server

stuzart (approved):

PR #1608, opened 2023-10-06
Extracted samples projects

fherreazcue:
Dissociates projects of samples and source data file.

### DIFF
--- a/lib/seek/samples/extractor.rb
+++ b/lib/seek/samples/extractor.rb
@@ -30,6 +30,7 @@ module Seek
 
             last_id = Sample.last.try(:id) || 0
             sample_type = samples.first.sample_type
+            project_ids = samples.first.project_ids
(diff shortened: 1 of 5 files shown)


lib/seek/samples/extractor.rb
```diff
@@ -38,6 +39,9 @@ def persist
             samples = Sample.where(sample_type: sample_type, title: samples.collect(&:title), contributor: contributor).where(
               'id > ?', last_id
             )
+            samples.each do |sample|
+              sample.project_ids = project_ids
+            end
```
stuzart, line 44: I'm not sure this is needed, if the project_ids for the first sample were correct, and the callbacks have been run with run_callbacks

It would be good to either add or expand a test to check `SetSubscriptionsForItemJob` are queued correctly

stuzart (changes requested):

stuzart (approved):

PR #1610, opened 2023-10-09
Experiment template make over

kdp-cloud:
This PR contains a a new feature where users can create new templates, based on existing templates.


app/views/templates/index.html.erb
somathias, line 1: This file should only contain the line
`<%= render :partial => "general/index" %>`
As it is now the title is displayed twice

image

kdp-cloud: Thanks for spotting that. Just pushed the changes.

app/views/templates/_form.html.erb
```diff
@@ -1,40 +1,66 @@
...
+                <div class="form-group">
+                    <%= f.hidden_field :parent_id %>
+                </div>
+
+            </div>
 
             <h3>Attributes</h3>
+            <div class="alert alert-warning alert-dismissible" role="alert">
+                <a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
+                <b class="text-danger">To ensure compliance with the original template, please do not modify the existing ISA-tags and attributes in the form below. You can always add new attribute.</b>
```
somathias, line 61: Typo. Should read either "You can always add new attributes." or "You can always add a new attribute."

app/views/templates/_form.html.erb
```diff
@@ -106,12 +134,23 @@
         $j('#attribute-table').on('change', '.sample-type-attribute-type', SampleTypes.attributeTypeChanged);
         $j('#attribute-table').on('change', '.controlled-vocab-selection', SampleTypeControlledVocab.controlledVocabChanged);
         $j('#attribute-table .sample-type-attribute-type').trigger("change", [false]);
-				$j('.templates').select2({theme: "bootstrap"});
-				initSelect2($j('.templates'), $j('#existing_templates'))
+        $j('.templates').select2({theme: "bootstrap"});
```
stuzart, line 137: would be good, for consistency and future maintainabiliy, if in the future this could be refactored to use the `objects_input.js` or better the `objects_input ` helper. But I don't think it should be a blocker for this PR

kdp-cloud: Good point. I made an issue (https://github.com/seek4science/seek/issues/1622) for this.

somathias (approved):

PR #1617, opened 2023-10-13
Datafile attribute relates samples and datafiles

### LINKED ISSUE #1534: Include in DF.related_samples those that have been linked via a Sample Type data file attribute
Currently the related samples only includes those that have originated from the DataFile via extraction.
Datafiles and Samples can also be linked via the Sample Type, "Registered DataFile" attribute but these are not included.

There is a request that these should be included - but is there a need to distinguish between the 2 cases? They are only related, and there are other cases where related items can come from multiple sources

fherreazcue:
Samples with data file attributes are now "related" to the data files.
- Sample's related data files include the originating data file and the data files listed in its attributes.
- Data file's related samples include the samples extracted from it and the samples that list it in their attributes.
Resolves #1534 .

### DIFF
--- a/app/models/data_file.rb
+++ b/app/models/data_file.rb
@@ -122,7 +122,8 @@ class DataFile < ApplicationRecord
   end
 
   def related_samples
-    extracted_samples
+    sample_ids = SampleResourceLink.where('resource_type': DataFile, 'resource_id': id).pluck('sample_id')
+    extracted_samples + Sample.where(id: sample_ids)
   end
 
   # Extracts samples using the given sample_type
--- a/app/models/data_file.rb
+++ b/app/models/data_file.rb
@@ -20,6 +20,8 @@ class DataFile < ApplicationRecord
 
   belongs_to :file_template
   has_many :extracted_samples, class_name: 'Sample', foreign_key: :originating_data_file_id
+  has_many :sample_resource_links, -> { where(resource_type: 'DataFile') }, class_name: 'SampleResourceLink', foreign_key: :resource_id
+  has_many :linked_samples, through: :sample_resource_links, source: :sample
 
   has_many :workflow_data_files, dependent: :destroy, autosave: true
   has_many :workflows, ->{ distinct }, through: :workflow_data_files
@@ -122,8 +124,7 @@ class DataFile < ApplicationRecord
   end
 
   def related_samples
-    sample_ids = SampleResourceLink.where('resource_type': 'DataFile', 'resource_id': id).pluck('sample_id')
-    extracted_samples + Sample.where(id: sample_ids)
+    extracted_samples + linked_samples
   end
 
   # Extracts samples using the given sample_type
--- a/app/models/data_file.rb
+++ b/app/models/data_file.rb
@@ -122,7 +122,7 @@ class DataFile < ApplicationRecord
   end
 
   def related_samples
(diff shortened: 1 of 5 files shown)


lib/tasks/seek_upgrades.rake
```diff
@@ -84,6 +85,20 @@ namespace :seek do
...
+    puts '... updating sample_resource_links for samples with data_file attributes...'
+    samples_updated = 0
+    disable_authorization_checks do
+      df_attrs = SampleAttribute.joins(:sample_attribute_type).where('sample_attribute_types.base_type' => Seek::Samples::BaseType::SEEK_DATA_FILE).pluck(:id)
+      samples = Sample.joins(sample_type: :sample_attributes).where('sample_attributes.id' => df_attrs)
+      samples.each do |sample|
+        sample.send(:update_sample_resource_links)
+        samples_updated += 1
+      end
+    end
```
stuzart, line 98: This looks more complicated than it should need to be. If necessary, it should be put in a class with tests.
Also looks like in terms or repeatability, each time it is run it will replace the resource_links with new ones, so should skip if they've already been created

test/unit/data_file_test.rb
```diff
@@ -489,4 +489,25 @@ class DataFileTest < ActiveSupport::TestCase
     assert data_file.respond_to?(:data_format_annotations)
     assert data_file.respond_to?(:data_type_annotations)
   end
+
+  test 'related samples include extracted samples and samples linked in attributes' do
+    project = FactoryBot.create(:project)
+    df1 = FactoryBot.create(:data_file)   # No samples
+    df2 = FactoryBot.create(:data_file)   # Extracted samples
+    df3 = FactoryBot.create(:data_file)   # Linked in sample's attributes only
+    df4 = FactoryBot.create(:data_file)   # Extracted samples and linked in sample's attributes
```
stuzart, line 498: use decriptive variable names instead of code comments

app/models/data_file.rb
```diff
@@ -122,7 +122,8 @@ def possible_sample_types(user = User.current_user)
   end
 
   def related_samples
-    extracted_samples
+    sample_ids = SampleResourceLink.where('resource_type': 'DataFile', 'resource_id': id).pluck('sample_id')
```
stuzart, line 125: you should be able to do this through a has_many association

stuzart (changes requested):

lib/tasks/seek_upgrades.rake
fherreazcue: I've added an if to only update the links when they are empty in [1fffe06](https://github.com/seek4science/seek/commit/1fffe069b9d89245c01726582d5f708b0d914a85).
I am failing to see what is so complicated... I am just getting a list of samples that have df attributes and updating their links... I thought this would be much more efficient than looping through all the sample types and then their samples...

test/unit/data_file_test.rb
fherreazcue: Resolved in [d93f6df](https://github.com/seek4science/seek/commit/d93f6dff068a259b4b0f59ce619ea5790d7a4645)

app/models/data_file.rb
fherreazcue: Resolved in [9bdc250](https://github.com/seek4science/seek/commit/9bdc25028fee8643082a76d41053465f27661353).

stuzart (approved):

PR #1643, opened 2023-11-03
Projects' add new dropdown

### LINKED ISSUE #1614: "Add new" button for projects only lists investigations
**Describe the bug**
"Add new" button on project overview page only lists "Add new investigation" in the drop down menu.
image

**Expected behavior**
All assets that can be linked to projects should be selectable from the drop down menu.

fherreazcue:
Includes all assets that can be linked to projects in the "add new" dropdown menu.
Resolves #1614.

### DIFF
--- a/app/controllers/concerns/legacy/workflow_support.rb
+++ b/app/controllers/concerns/legacy/workflow_support.rb
@@ -11,7 +11,7 @@ module Legacy
     def create_ro_crate
       @crate_builder = Legacy::WorkflowCrateBuilder.new(legacy_ro_crate_params)
       @workflow.workflow_class = @crate_builder.workflow_class = WorkflowClass.find_by_id(params[:workflow_class_id])
-      @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids]
+      @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids] || []
       blob_params = @crate_builder.build
       @content_blob = ContentBlob.new(blob_params)
 
@@ -32,7 +32,7 @@ module Legacy
         if handle_upload_data && @workflow.content_blob.save
           @content_blob = @workflow.content_blob
           @workflow = workflow
-          @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids]
+          @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids] || []
           if extract_metadata(@content_blob)
             format.html { render :provide_metadata }
           else
--- a/app/controllers/concerns/legacy/workflow_support.rb
+++ b/app/controllers/concerns/legacy/workflow_support.rb
@@ -11,6 +11,7 @@ module Legacy
     def create_ro_crate
       @crate_builder = Legacy::WorkflowCrateBuilder.new(legacy_ro_crate_params)
       @workflow.workflow_class = @crate_builder.workflow_class = WorkflowClass.find_by_id(params[:workflow_class_id])
+      @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids]
       blob_params = @crate_builder.build
       @content_blob = ContentBlob.new(blob_params)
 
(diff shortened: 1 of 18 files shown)


app/controllers/concerns/legacy/workflow_support.rb
```diff
@@ -11,6 +11,7 @@ module WorkflowSupport
     def create_ro_crate
       @crate_builder = Legacy::WorkflowCrateBuilder.new(legacy_ro_crate_params)
       @workflow.workflow_class = @crate_builder.workflow_class = WorkflowClass.find_by_id(params[:workflow_class_id])
+      @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids] || []
```
stuzart, line 14: I don't think this is in the right place, it doesn't belong as part of `create_ro_crate `and should have been set already.
I think you just need to add `workflow[project_ids]` to the permitted params in legacy_ro_crate_params and also possibly update `legacy_set_workflow`

for what reason do we need to handle `params[:project_ids]`, rather than it always being `workflow[project_ids]` ?

app/controllers/concerns/legacy/workflow_support.rb
```diff
@@ -31,6 +32,7 @@ def create_content_blob
         if handle_upload_data && @workflow.content_blob.save
           @content_blob = @workflow.content_blob
           @workflow = workflow
+          @workflow.project_ids |= params[:workflow] ? params[:workflow][:project_ids] : params[:project_ids] || []
```
stuzart, line 35: same as for `create_ro_crate`

stuzart (changes requested):
I think it would also be good if @fbacall could check the workflow bits

app/controllers/concerns/legacy/workflow_support.rb
fbacall: Agree with @stuzart here. Also for the actions in WorkflowsController.

It's possible to do it all via the params, but unfortunately you have to do something slightly different in each case.

app/controllers/sample_types_controller.rb
```diff
@@ -28,7 +28,8 @@ def show
   def new
     @tab = 'manual'
 
-    @sample_type = SampleType.new
+    attr = params["sample_type"] ? send("sample_type_params") : {}
```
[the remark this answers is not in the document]
fherreazcue: Resolved in [e66050c](https://github.com/seek4science/seek/pull/1643/commits/e66050c89da969b540eb9d964105fc6adc7c4411).

app/helpers/multi_step_wizard_helper.rb
```diff
@@ -46,6 +46,9 @@ def forward_params(key)
       if params[key][:publication_ids]
         html << hidden_field_tag('data_file[publication_ids][]', params[key][:publication_ids])
       end
+      params[key][:project_ids]&.each do |p|
+        html << hidden_field_tag("#{key}[project_ids][]", p)
```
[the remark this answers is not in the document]
fherreazcue: Resolved in [e66050c](https://github.com/seek4science/seek/pull/1643/commits/e66050c89da969b540eb9d964105fc6adc7c4411).

app/views/investigations/_form.html.erb
```diff
@@ -42,12 +42,3 @@
 
 <%= form_submit_buttons(@investigation) %>
 
-<script>
```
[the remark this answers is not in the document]
fherreazcue: You are right, this broke the button you mention... I rolled it back in [564ebb6](https://github.com/seek4science/seek/pull/1643/commits/564ebb6cc46c5ca2a9d4d43bc3160a862c52aa63).

app/controllers/concerns/legacy/workflow_support.rb
fherreazcue: See 614e01f.

app/controllers/concerns/legacy/workflow_support.rb
```diff
@@ -46,9 +46,11 @@ def create_content_blob
     private
 
     def legacy_ro_crate_params
-      params.require(:ro_crate).permit({ workflow: [:data, :data_url, :make_local_copy] },
+      l_params = params.require(:ro_crate).permit({ workflow: [:data, :data_url, :make_local_copy]},
                                        { abstract_cwl: [:data, :data_url, :make_local_copy] },
                                        { diagram: [:data, :data_url, :make_local_copy] })
+      l_params[:workflow][:project_ids] = params.dig(:workflow, :project_ids) || []
+      l_params
```
stuzart, line 53: this looks a bit weird. Can you not just add `[:project_ids]` to the workflow params, 
i.e. `.permit({ workflow: [:data, :data_url, :make_local_copy, project_ids]}`

stuzart (changes requested):

stuzart (approved):

fbacall (approved):

Issue #1632, opened 2023-10-20
Sample attribute types still mistakenly includes Ontology


stuzart:
Ontology just duplicates Controlled Vocabulary

floradanna:
what could be a quick fix for that?

does it relate to #802 ?

stuzart:
I've just been looking at it. I will need to find all sample_attributes associated with the Ontology attribute type, and moved them to the "Controlled Vocabularly" attribute type, and then delete the Ontology one.

I'm not sure #802 is relevant any more. A different flag is now used to determine whether a CV can accept new terms, regardless of whether it was populated from an ontology or not. ( `.custom_input?` ). It's set on the CV, but we have discussed that it would be better on the sample attribute as it depends on how the CV is used

floradanna:
does it mean that the attribute type Ontology will not exist anymore?

@kdp-cloud we use attribute type = ontology in our json templates. Will this change affect the templates? 
Initially, we made the DT so that
- attribute type = CV --> accept only terms from the list
- attribute type = ontology --> ontology terms are suggested, but user can type any free text
I think this behaviour is already compromised in the DT.
Shall we address this issue step-by-step and work on short-term and long-term solution?
@stuzart please, let us know what you think.

About `.custom_input?`, I think I would need more information, but we would want to avoid:
- that CV List get edited by end-users at any moment
- confusion between attributes that enforce CV List and attributes that don't without any clear distinction or reason

stuzart:
Both Controlled Vocabularly and Ontology attribute types are exactly the same, and are defined the same.
There was only ever a hard-coded hack that marked some controlled vocabs as allowing free text based on an ontology flag in the html markup, but I don't think this was set according to the attribute type. It led to some bugs and therefore was changed to use an explicit `.custom_input` regardless of where the CV came from.

As you can see here, both Controlled Vocab and Ontology have the same definition and base type, just a different title:
CV - https://github.com/seek4science/seek/blob/main/db/seeds/007_sample_attribute_types.seeds.rb#L67-L68
Ontology - https://github.com/seek4science/seek/blob/main/db/seeds/007_sample_attribute_types.seeds.rb#L82-L83

both take you to the same pages and have the same behaviour.

The Ontology attribute type is causing confusion. We don't support ontologies, but only support populating a Controlled Vocab with terms derived from an Ontology (there using a subset of an ontology as a CV, but not any other ontology properties)

PR #1647, opened 2023-11-09
Remove ontology attribute type

stuzart:
Remove '_Ontology_' `SampleAttributeType`, which was really a duplicate of '_Controlled Vocabulary_'

required removing some legacy javascript behaviour that filtered the list of CV's available, but based on `.custom_input?` rather than whether they are ontology based. 
An upgrade task migrates existing sample attributes linked to '_Ontololgy_' to '_Controlled Vocabularly_'

* fix for  #1632

### DIFF
--- a/app/assets/javascripts/sample_types.js
+++ b/app/assets/javascripts/sample_types.js
@@ -72,12 +72,10 @@ var SampleTypes = {
     attributeTypeChanged: function (e, resetSelection=true) {
         //check if it is a controlled vocab, and change the state of the controlled vocab selector if need be
         var use_cv = $j(this).find(':selected').data('use-cv');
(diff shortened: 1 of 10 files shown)


kdp-cloud (approved):
There is some old 'hash rocket' syntax in [app/views/sample_types/_sample_attribute_form.html.erb](https://github.com/seek4science/seek/pull/1647/files#diff-15f759ac92df0b507336c36fc5e1a74de9905ffc263dba98bc6cc0d78fd15a49) and [app/views/templates/_template_attribute_form.erb](https://github.com/seek4science/seek/pull/1647/files#diff-f3f00cd514ded118fe921a38bab9a104f72e63b693c461a20303d683a146998f) you might consider updating to the new JSON syntax.
Other than that, no comments.

PR #1669, opened 2023-11-21
Fix isa exporter permissions

kdp-cloud:
This PR implements the following:
- Checks whether the user has at least viewing permission to the ISA
  - A user needs viewing permission for the investigation
  - A user needs viewing permission for the studies.
  - A user needs viewing permission for all the assays in an assay stream (guarantees the linkage)
- Checks whether the user has at least viewing permission to the samples within that ISA.
  - If the user does not have the right permissions, it will hide the sample information.

### DIFF
--- a/lib/isa_exporter.rb
+++ b/lib/isa_exporter.rb
@@ -323,11 +323,19 @@ module IsaExporter
         sample_type.sample_attributes.select { |sa| sa.isa_tag&.isa_source_characteristic? }
 
       sample_type.samples.map do |s|
-        {
-          '@id': "#source/#{s.id}",
-          name: s.get_attribute_value(with_tag_source),
-          characteristics: convert_characteristics(s, with_tag_source_characteristic)
-        }
+        if s.can_view?(@current_user)
+          {
+            '@id': "#source/#{s.id}",
+            name: s.get_attribute_value(with_tag_source),
+            characteristics: convert_characteristics(s, with_tag_source_characteristic)
+          }
+        else
+          {
+            '@id': "#source/HIDDEN",
+            name: 'Hidden Sample',
+            characteristics: []
+          }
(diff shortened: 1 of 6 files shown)


lib/isa_exporter.rb
```diff
@@ -585,6 +640,18 @@ def group_samples_by_input_and_parameter_value(sample_type)
       grouped_samples.transform_keys { |key| group_id(key) }
     end
 
+    def process_sequence_input(inputs, type)
+      input_ids = inputs.map { |input| input[:id] }
+      authorized_sample_ids = Sample.where(id: input_ids).select { |s| s.can_view?(@current_user) }.map(&:id).compact
```
stuzart, line 645: you could use ...`Sample.where(id: input_ids).authorized_for(:view, @current_user)`... which will filter in a single query. Might not matter here, but can make a difference where there are large number and avoids looping over.
Not a blocker though

lib/isa_exporter.rb
```diff
@@ -623,8 +694,17 @@ def convert_parameter_values(sample_group_hash, isa_parameter_value_attributes)
       end
     end
 
-    def extract_sample_ids(obj, type)
-      Array.wrap(obj).map { |item| { '@id': "##{type}/#{item[:id]}" } }
+    def extract_sample_ids(input_obj_list, type)
+      sample_ids = input_obj_list.map { |io| io[:id] }
+      authorized_sample_ids = Sample.where(id: sample_ids).select { |sample| sample.can_view?(@current_user) }.map(&:id)
```
stuzart, line 699: again, authorized_for might be better

stuzart (approved):
a few cases where using `.authorized_for()` would be more efficient for filtering authorized items, but only an issue if expecting a large number

Issue #1648, opened 2023-11-09
move custom_input? from Controlled Vocab to Sample attribute


stuzart:
Rather than being tied to a particular CV, instead define custom_input (i.e allows unrestricted terms) to the attribute where the CV is being used. This means that the same CV can be used, but allowing custom input is dependent on how it is used.

Needs a UI option to set when defining an attribute type as CV

Migration needs to move the value from ControlledVocab to SampleAttribute before deleting the column, so cannot be done as an upgrade task

PR #1672, opened 2023-11-21
Move custom input to attribute UI part plus some refactoring

stuzart:
Updated the UI to handle setting on a sample attribute whether is allows free text.

Also some refactoring around the attribute-type-handlers so that they take the attribute in the initializer.
Also some refactoring to simplify how the handlers are used, so that they are used via the attribute rather than sample type. This was mainly being done through delegation anyway so makes things simpler

* part of #1648

Final step to do is to change the "AttributeTypeHandlers" into just "AttributeHandlers" - they currently take the base_type and attribute, but it is now possible to get the base_type via the attribute. This will mostly require updating tests, but will be done as a separate PR to avoid large PRs


kdp-cloud (approved):
Pushed a commit with changes for the dynamic table.

PR #1680, opened 2023-11-28
Issue 1676 fix cv list in dynamic table

kdp-cloud:
Until now CV lists were not available to use in the dynamic table and were not considered in the development of all features around it.

- Use of CV Lists in the dynamic table, shown as select2
- Modified the download to spreadsheet feature to accommodate CV Lists
- Modified the upload to spreadsheet feature to handle and validate CV List entries


app/views/isa_assays/_assay_samples.html.erb
```diff
@@ -1,3 +1,9 @@
+<%
+=begin%>
+ The dynamic table where users can interact with
+<%
+=end%>
+
```
stuzart, line 6: for small comments you can use <%# comment %>

app/views/single_pages/download_samples_excel.axlsx
```diff
@@ -2,7 +2,7 @@ require 'caxlsx'
 require 'uuid'
 
 #####################################
-debug = false
+debug = true
```
stuzart, line 5: is this intended ?

stuzart (approved):
added a couple of comments but no blockers.
some of the javascript is pretty complex and I can't say I didn't miss any errors but I got the gist of it. I wonder if some of it, like building the dynamic table, might be simpler done as an ajax call and then avoid some of the special string TOKENS that need replacing with the values - but just an idea for some future refactoring

app/views/isa_assays/_assay_samples.html.erb
```diff
@@ -1,8 +1,4 @@
-<%
-=begin%>
- The dynamic table where users can interact with
-<%
-=end%>
+<%= # The dynamic table where users can interact with %>
```
stuzart, line 1: the `#` should directly follow the `<%` with no `=` or space
e.g. `<%# The dynamic table where users can interact with %>`

PR #1685, opened 2023-11-29
Auto approve programmes and projects in site-managed programme

fbacall:
Adds admin settings to:
* automatically activate programmes.
* automatically approve project creation requests in site-managed programme.

(also adds mailer previews in development environment: http://localhost:3000/rails/mailers)


stuzart (changes requested):
If requesting a new Programme and Project together and `auto_activate_programmes` is `true` then it will still expect approval instead of automatically activating the new programme.

stuzart (approved):

Issue #1687, opened 2023-11-29
Show the admin job queue and queued items in scrollable panes


stuzart:
to avoid having to scroll down through a long list to get to the next section. 

The Job queue, and the Queued items should be in separate pane, not all in 1

PR #1694, opened 2023-12-11
Show admin job queue in scrollable panes 1687

stuzart:
Both the queue of delayed jobs and the queue of items are displayed in scrollable panes.
Avoid the problem of trying to find the seperation between the 2 tables, to check the queued items, when there is a long backlog, and also avoid having to scroll far down the page.

* Fix for #1687


whomingbird (approved):

whomingbird:
UI and code are both OK.

PR #1697, opened 2023-12-13
Import projects from Data Management Plan files

Menjabin:
Adds a workflow for importing projects from Data Management Plan files that follow the RDA DMP Common Standard. This format is used by e.g. ELIXIR's Data Stewardship Wizard, and it would greatly simplify the process of publishing projects from there to a SEEK instance if it could be done by a simple file upload rather than by manually entering the data.

This solution reads the data about a project from a DMP file, and this is used to create a new project. All contributors to the project are also read from the file, and user accounts are created and added as members of the newly created project.
The option to import a project is currently only presented to the user on both the "new project" and "guided create" pages. It is not accessible through the main navbar.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -318,12 +318,17 @@ SEEK::Application.routes.draw do
       get :request_institutions
       get :guided_join
       get :guided_create
+      get :guided_import
       post :request_join
       post :request_create
+      post :request_import
       get :administer_create_project_request
+      get :administer_import_project_request
       post :respond_create_project_request
+      post :respond_import_project_request
       get :project_join_requests
       get :project_creation_requests
+      get :project_importation_requests
       get  :typeahead
     end
     member do
@@ -737,7 +742,7 @@ SEEK::Application.routes.draw do
       post :template_attributes
     end
     collection do
-      post :filter_isa_tags_by_level
+post :filter_isa_tags_by_level
       get :task_status
       get :default_templates
       post :populate_template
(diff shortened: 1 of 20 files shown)


stuzart:
Hi,
There is something wrong with the PR and cannot review or merge. It contains many changes to many files, many of them reverting recent changes. I've tried looking at the commit itself by cloning and checking outside of github - and it is the same there too.
It's possible that resyncing with the latest main branch may solve things, but otherwise it will require a new PR that contains only the changes for the new feature.

Menjabin:
My apologies,
I have updated the branch so that only my changes remain.

stuzart (changes requested):
it lacks any functional or unit tests, and due to being quite complex and involved it really needs.
Without them it is difficult to review the pull request works, and also there is high potential for things to break without knowing with future changes

Menjabin:
I have now added tests for the new functionality

stuzart (approved):

config/routes.rb
```diff
@@ -737,7 +742,7 @@
       post :template_attributes
     end
     collection do
-      post :filter_isa_tags_by_level
+post :filter_isa_tags_by_level
```
kdp-cloud, line 745: Missing indentation?

kdp-cloud (approved):
There's a missing indentation in `routes.rb` on line 745.
Other than that, no comments!

config/routes.rb
stuzart: we can fix that after it's merged

stuzart:
thanks for the contribution!

Issue #1657, opened 2023-11-14
relation between P-ISA levels


floradanna:
[RAD 5 - Edpi ID 5.2](https://docs.google.com/document/d/1FoRDCDlbmeNJgF0gJ4M5JJ-QjNyzhCNe42DxJ2rStxI/edit#heading=h.u3jbo9p4b1dp)

While in default SEEK user can edit the associated higher level in ISA structure (e.g., assay to study etc), this in currently prevented in DataHub (from Single Page).

Investigation to Project: see #1351 . 
- allowing to change the associated project **should not** have any impact on ISA
- at first, let's keep the feature that prevent user from changing associated project for ISA-JSON compliant Investigation

Study to Investigation
- allowing to change the associated investigation **should not** have any impact on ISA
- at first, let's keep the feature that prevent user from changing associated Investigation for ISA-JSON compliant Study

Assay to Study
- allowing to change the associated study **can** create confusion because of sources and samples at the study level used in specific assay
- let's keep the feature that prevent user from changing associated study for ISA-JSON compliant Assay

Sample Type 
- ST from Single Page (ISA-JSON compliant) are hidden from the user via GUI to prevent any changes directly. Feature to keep for now. In the future, clicking on ISA-JSON compliant ST could bring the user to the related Assay or Study.
- Do not allow user to edit ISA-JSON compliant ST associated project

Samples
- BUG (probably): user can edit ISA-JSON compliant sample via default view and associated to a different project directly. This must be prevented for now. Deeper analysis is needed.

Issue #1570, opened 2023-08-21
Add a tag to identify I-S-A that are ISA-JSON compliant


floradanna:
Currently, the sole distinction between Creating/Editing a "vanilla seek ISA" and an "ISA-JSON compliant experiment" lies in the source of the request: whether it originates from the default view or a single page.

This is a week system that causes many conflicts, such as allowing users 
- to edit "ISA-JSON compliant experiment" using Editing option available from default view and vice verse
- link "vanilla seek study or assay" to Investigation or Study containing "ISA-JSON compliant experiment", therefore breaking ISA-JSON export at investigation level.

A more optimal solution would involve adding an option (or a tag, a parameter, etc.) during the Investigation creation process when SP advanced is enabled. This option would allow the user to decide whether the Investigation will exclusively contain "vanilla seek ISA" or "ISA-JSON compliant experiments".

Then, the different GUI to create/edit "vanilla seek ISA" or "ISA-JSON compliant experiments" could be shown to the user based on this information.

floradanna:
[RAD Action_1 - Epic 1.8](https://docs.google.com/document/d/1ByETT3mssb_Ng4VYPwSw7MlxfvRefKqsCcCZvgtUCXY/edit#heading=h.jd63gy2wgf0s)

somathias:
This can be done by adding a boolean flag to investigations, called eg. "is_fully_ISA_compliant". 
As Studies and Assays always link to an investigations, they "inherit" the property from their top-level investigation.

If ISA-JSON compliance is enabled, the create investigation form should display a checkbox for tagging the investigation as "fully_ISA_compliant".

floradanna:
Could we implement this in a way that "fully ISA-JSON compliance" would be the ONLY way to create ISA in a specific instance? So, maybe having some instance-specific configuration that makes "fully ISA-JSON compliance" the default.

Instance admin could then configure the instance in 3 possible ways:
- default ISA only
- default ISA and "fully ISA-JSON compliant ISA"
- "fully ISA-JSON compliant ISA" only

Let me know if what I mean is not clear.

somathias:
Assuming that all three configuration options are desired (as opposed to just having either "default ISA only" or "fully ISA-JSON compliant only"):

this could be implemented using an additional subflag/checkbox in the admin panel "Allow only full ISA-JSON compliance", that becomes available when checking "Compliance to ISA-JSON schemas enabled". 
[...]

PR #1698, opened 2023-12-13
Identify ISA-JSON compliant investigations

### LINKED ISSUE #1355: Edit ISA Study returns 500
**Describe the bug**
It's possible to edit the study under "normal view" but trying to edit the study under "Single Page" `Edit ISA-Study` returns error code 500. `Edit ISA-Study`contains the sample source which we want to access. 

**To Reproduce**

-1 Under Single Page view of Study.
-2 Click edit ISA Study under Actions

**Expected behavior**
Able to edit the 

**Screenshots**
https://files.slack.com/files-pri/T02HT5X58-F04P822U38D/screenshot_2023-02-13_at_13.37.35.png

**Desktop (please complete the following information):**
MacOS
Tested on both 
Firefox 109.0.1  and 
Chrome Version 109.0.5414.119 

Log output:

```
E, [2023-02-13T11:21:36.236968 #766] ERROR -- : [246142700da6ba47c0efa085e0dc70b7] ERROR - NoMethodError (undefined method `can_perform?' for nil:NilClass)
E, [2023-02-13T11:21:36.237100 #766] ERROR -- : [246142700da6ba47c0efa085e0dc70b7] Sending execption ERROR - NoMethodError (undefined method `can_perform?' for nil:NilClass)
I, [2023-02-13T11:21:36.239611 #766]  INFO -- : [246142700da6ba47c0efa085e0dc70b7] method=GET path=/isa_studies/2/edit format=html controller=IsaStudiesController action=edit st\
atus=500 duration=41.94 view=1.97 db=3.99 time=1045999.23 user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safa\
ri/537.36

```
Encountered by @parul12

somathias:
Adds a column to the investigations table to check if investigations are ISA-JSON compliant. Derived studies and assays inherit the flag from their parent investigation if they have the correct sample types associated. Furthermore, makes a initial simple attempt to entangle functionality related to handling ISA-JSON compliant elements from the Single Page view. 

More specifically, if the ISA-JSON compliance setting is enabled in the admin panel the following functionality is available:
- the form for creating investigations includes a check box for marking the investigation as ISA-JSON compliant.
- the behavior of buttons for handling Investigations, Studies and Assays depends on whether they are ISA-JSON compliant or not. This functionality is hence no longer dependent on viewing the Study in Single page or default view. ISA-JSON compliant investigations e.g. always display the button "Design Study" and non-ISA-JSON compliant investigations can be managed in SP view in the default way.
- ISA-JSON compliant Studies and Assays display the "Study Design"/"Assay Design" tabs also in default view. ~However, the dynamic table can currently only be populated when in Single Page view. In default view it remains empty.~
- Projects, Investigations, Studies and Assays all have buttons to switch between Single Page and default view and back.
- Once a Study has been associated with an Investigation, the is-ISA-JSON-compliant flag cannot be changed. 
- Default Studies/Assays that are created through the Create menu can only be linked to non-ISA-JSON compliant Investigations/Studies.

This PR addresses #1570, fixes #1355 and relates to #1657.


kdp-cloud:
Single Pages controller also the `export_isa function` [here](https://github.com/seek4science/seek/blob/bc101acd945425a0d22b569ee88dbc10cd575bb0/app/controllers/single_pages_controller.rb#L52-L63). You should delete this.

config/routes.rb
```diff
@@ -405,6 +405,7 @@
     resources :people, :programmes, :projects, :assays, :studies, :models, :sops, :workflows, :data_files, :publications, :documents, only: [:index]
     member do
       get :export_isatab_json
+      get :export_isa, action: :export_isa
```
kdp-cloud, line 408: You should delete the route in the single_pages resource since it's not used anymore:
https://github.com/seek4science/seek/blob/bf3200c8d4c8c9b921269b5b233a76299394dc79/config/routes.rb#L754

app/views/assays/show.html.erb
```diff
@@ -124,7 +134,7 @@
   <%= tab_pane('related-items') do %>
     <%= render partial: 'general/items_related_to', object: @assay %>
   <% end %>
-	<% if Seek::Config.isa_json_compliance_enabled && displaying_single_page? %>
+	<% if Seek::Config.isa_json_compliance_enabled && @assay.is_ISA_JSON_compliant %>
```
kdp-cloud, line 137: I don't know if you were planning on fixing it at a later stage but I just want to mention here that this will not work without modifying the dynamic table logic.
It requires a selected 'Samples table', which represents a sample type, from the treeview.

config/routes.rb
somathias: Good catch - fixed in [60fcdde](https://github.com/seek4science/seek/pull/1698/commits/60fcddee7fdfdeac3ece58fc81c8df8b8798b260)

app/views/assays/show.html.erb
somathias: Yes, we are aware that this requires future modification of the dynamic table logic to work (and that it hence currently does not).

kdp-cloud (commented):
Would be a good idea to add some tests:

- [x] Unit test for studies testing the ISA-JSON compliance. 
- [ ] Unit test for assays testing the ISA-JSON compliance. 
- [ ] Functional test testing the button text when an investigation is considered ISA-compliant.
- [ ] Functional test testing the button text when a study is considered ISA-compliant.
- [ ] Functional test testing the button text when an assay is considered ISA-compliant.

stuzart (changes requested):
a minor comment, but important, but to conform with the ruby code style guide, method names, and also database attributes, shouldn't have capitals. So
     `is_ISA_JSON_compliant`
should be
     `is_isa_json_compliant`
mysql itself isn't case senstive for column names, but rails will automaticaly be using that for it's attribute name - so will require either a new migration to change the column name or can be rolled back (`rake db:rollback`) and then change the existing migration before running the migration again

where calling `is_isa_json_compliant`, because it's boolean it should end with a ?, i.e `is_isa_json_compliant?` - which really helps interpreting the code

somathias:
I have now renamed the database attribute by adding a new migration and added a ? to the boolean method name to adhere to the ruby code style convention. See commits [c7a030b](https://github.com/seek4science/seek/pull/1698/commits/c7a030b8a6c68679be8a42004c9e13b113fdddaa) and [2a9db2b](https://github.com/seek4science/seek/pull/1698/commits/2a9db2b909b24b1aa65a68bc5b72ad81682092eb).

stuzart (approved):

kdp-cloud (approved):

PR #1708, opened 2024-01-02
Add attributes to existing sample types

kdp-cloud:
- Add policies and sharing permissions to sample types
- Making existing sample types more editable
- Simplifying assays and studies creation forms.


app/controllers/sample_types_controller.rb
```diff
@@ -133,43 +145,46 @@ def filter_for_select
...
-  def batch_upload
-
-  end
+  def batch_upload; end
 
   private
 
+  def update_sample_json_metadata
+    # UpdateSampleMetadataJob.new(@sample_type).queue_job
+    UpdateSampleMetadataJob.new(@sample_type).perform_now
```
stuzart, line 154: should be `UpdateSampleMetadataJob.perform_later(@sample_type)` and only triggered when neccessary

stuzart: since this is quite common across several controllers, I'm wondering if it should go in a module and include checks for whether sample type attributes have changed in a particular way

app/jobs/update_sample_metadata_job.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateSampleMetadataJob < ApplicationJob
+  queue_with_priority 1
+  queue_as QueueNames::SAMPLES
+  def perform(sample_type)
+    Seek::Samples::MetadataUpdater.new(sample_type).update_metadata
```
stuzart, line 5: this is going to need some profiling and scalability testing with a sample type linked to large set of samples. I'll get some metrics from fairdomhub about what is a our current maximum

app/views/assets/_manage_form.html.erb
```diff
@@ -9,10 +10,19 @@
...
+      <h4>ISA Compliant structure</h4>
+      <p>
+        This <%= "#{t(resource_class)}".downcase.html_safe%> is part of an ISA compliant structure. The permissions of all sample types linked to it will be equally updated.
+      </p>
+    </div>
+    <% end %>
 
-<% end -%>
\ No newline at end of file
+  <%= form_submit_buttons(resource, { validate:false, update_sample_types: true }) %>
```
stuzart, line 26: I don't see where `update_sample_types` param is used, and shouldn't be set to true in all cases

app/views/sample_types/_form.html.erb
```diff
@@ -18,6 +18,7 @@
     </div>
 
     <%= render partial: "projects/project_selector", locals: { resource: @sample_type } -%>
+    <%= render partial: 'assets/manage_specific_attributes', locals:{f:f} if show_form_manage_specific_attributes? %>
```
stuzart, line 21: for consistency, should use `assets/asset_form_bottom` with the relevant options passed

app/views/sample_types/_form.html.erb
```diff
@@ -51,13 +50,11 @@
                                                                            display_isa_tag: false } %>
           <% end %>
 
-          <% unless @sample_type.uploaded_template? || !@sample_type.editing_constraints.allow_new_attribute? %>
```
stuzart, line 54: check for @sample_type.uploaded_template? should remain for now, but might be revisited as part of #1766

db/schema.rb
```diff
@@ -1715,7 +1715,9 @@
     t.text "description"
     t.integer "isa_tag_id"
     t.boolean "allow_cv_free_text", default: false
+    t.integer "template_attribute_id"
```
stuzart, line 1718: not part of the migration, looks like contamination from switching branches

db/schema.rb
```diff
@@ -1715,7 +1715,9 @@
...
     t.index ["sample_type_id"], name: "index_sample_attributes_on_sample_type_id"
+    t.index ["template_attribute_id"], name: "index_sample_attributes_on_template_attribute_id"
```
stuzart, line 1720: not part of the migration, looks like contamination from switching branches

db/schema.rb
```diff
@@ -2101,6 +2116,8 @@
     t.string "pid"
     t.boolean "allow_cv_free_text", default: false
     t.integer "linked_sample_type_id"
+    t.integer "parent_attribute_id"
+    t.index ["parent_attribute_id"], name: "index_template_attributes_on_parent_attribute_id"
     t.index ["template_id", "title"], name: "index_template_id_asset_id_title"
```
stuzart, line 2121: not part of the migration, looks like contamination from switching branches

stuzart (changes requested):
A partial review, to get started with.

PR #1716, opened 2024-01-10
Add rake upgrade task for recognising ISA JSON compliant investigations

kdp-cloud:
Upgrade task that sets the is_isa_json_compliant field to true of  all investigations that are linked to studies with sample types.

### DIFF
--- a/lib/tasks/seek_upgrades.rake
+++ b/lib/tasks/seek_upgrades.rake
@@ -15,6 +15,7 @@ namespace :seek do
     rename_registered_sample_multiple_attribute_type
     remove_ontology_attribute_type
     db:seed:007_sample_attribute_types
+    recognise_isa_json_compliant_items
   ]
 
   # these are the tasks that are executes for each upgrade as standard, and rarely change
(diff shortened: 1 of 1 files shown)


somathias (approved):

lib/tasks/seek_upgrades.rake
```diff
@@ -156,6 +157,21 @@ namespace :seek do
...
+    puts '... searching for ISA compliant investigations'
+    investigations_updated = 0
+    disable_authorization_checks do
+      isa_json_compliant_studies = Study.all.select { |study| study.sample_types.any? }
+      isa_json_compliant_studies.each do |study|
+        inv = study.investigation
+        inv.update(is_isa_json_compliant: true)
+        inv.save
+        investigations_updated += 1
+      end
```
stuzart, line 170: to be more effecient and repeatable, you could limit the studies to only those linked to an investigation where is_isa_json_compliant is false. Something like `Study.join(:investigation).where('study.investigation.is_isa_json_compliant = ?', false).select ...`  (probably not quite right, but you get the idea).
Also, it looks like it is updating the same investigation multiple times if it has multiple studies.

You should use `inv.update_column(:is_isa_json_compliant, true)` to avoid triggering callbacks and and changing timestamps

PR #1722, opened 2024-01-17
Implement assay streams for single page

kdp-cloud:
- Introduction of Assay stream `AssayClass`.
- Implement usage of Assay streams to hold ISA Assays
- Adapt ISA JSON exporters to handle assay streams


app/models/assay_class.rb
```diff
@@ -1,31 +1,38 @@
...
-
-  #this returns an instance of AssayClass according to one of the types "experimental" or "modelling"
-  #if there is not a match nil is returned
-  def self.for_type type
-    keys={"experimental"=>"EXP","modelling"=>"MODEL"}
-    return AssayClass.find_by(key: keys[type])
+  # this returns an instance of AssayClass according to one of the types "experimental", "modelling" or "assay_stream"
+  # if there is not a match nil is returned
+  def self.for_type(type)
+    keys = { "experimental": 'EXP', "modelling": 'MODEL', "assay_stream": 'ASS' }
```
fbacall, line 5: Can we go with a different key for assay stream? e.g. `STREAM`

app/models/assay_class.rb
```diff
@@ -1,31 +1,38 @@
...
+    for_type('experimental')
   end
 
   def self.modelling
-    self.for_type('modelling')
+    for_type('modelling')
+  end
+
+  def self.assay_stream
+    for_type('assaystream')
```
fbacall, line 18: should be `assay_stream`?

app/models/assay_class.rb
fbacall: Also I think this mapping should be moved out of this method into a constant.

app/models/assay_class.rb
```diff
@@ -1,31 +1,38 @@
...
 
+  def is_assay_stream?
+    key == 'ASS'
+  end
+
   # for cases where a longer more descriptive key is useful, but can't rely on the title
   #  which may have been changed over time
   def long_key
-    {'EXP'=>'Experimental Assay','MODEL'=>'Modelling Analysis'}[key]
+    { 'EXP': 'Experimental Assay', 'MODEL': 'Modelling Analysis', 'ASS': 'Assay Stream' }[key.to_sym]
```
fbacall, line 36: See above re: constant (although this would require the mapping to be inverted)

app/views/assays/_buttons.html.erb
```diff
@@ -1,4 +1,14 @@
...
+<% assay_word ||=
+  if item.is_modelling?
+    t('assays.modelling_analysis')
+  elsif item.is_assay_stream?
+    t('assays.assay_stream')
+  elsif Seek::Config.isa_json_compliance_enabled && item.is_isa_json_compliant?
+    t('isa_assay')
+  else
+    t('assays.assay')
+  end
```
fbacall, line 10: I think this could be simplified to check `t("assays.#{item.assay_class.long_key}")` or something for the non-ISA assay cases

lib/tasks/seek_upgrades.rake
```diff
@@ -176,6 +178,48 @@ namespace :seek do
...
+    puts '... Organising isa json compliant assays in assay streams'
+    assay_streams_created = 0
+    disable_authorization_checks do
+      # find assays linked to a study through their sample_types
+      # Should be isa json compliant
+      # Shouldn't already have an assay stream (don't update assays that have been updated already)
+      # Previous ST should be second ST of study
+      first_assays_in_stream = Assay.joins(:sample_type)
+        .select(&:is_isa_json_compliant?)
+        .select { |a| a.assay_stream_id.nil? && (a.previous_linked_sample_type == a.study.sample_types.second) }
```
fbacall, line 191: I would try and narrow down this query if you can to avoid needing to load every assay into memory.

I think you most of this filtering at the database level by doing something like:
`Assay.joins(:sample_type, study: :investigation).where(assay_stream_id :nil, investigation: { is_isa_json_compliant: true })`

app/views/assays/_buttons.html.erb
kdp-cloud: It can but the button will say 'New __Experimental assay__ based on this one' instead of 'New __Assay__ based on this one'.

lib/treeview_builder.rb
```diff
@@ -14,9 +14,22 @@ def build_tree_data
     investigation_items = []
 
     @project.investigations.map do |investigation|
-      investigation.studies.map do |study|
-        assay_items = study.assays.map { |assay| build_assay_item(assay) }
-        study_items << build_study_item(study, assay_items)
+      if investigation.is_isa_json_compliant?
+        investigation.studies.map do |study|
+          assay_stream_items = study.assays.select { |assay| assay.is_assay_stream? }.map do |assay_stream|
```
fbacall, line 19: You can use your `study.assay_streams` method here

lib/seek/isa/assay_class.rb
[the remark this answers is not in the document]
fbacall: Or if you do want to use constants instead of strings, just put them in AssayClass e.g. `AssayClass::EXP`, to save some typing.

fbacall (approved):

stuzart (approved):

floradanna:
[testing](https://docs.google.com/document/d/1sVsfUg_mIv2tBIQ2AJlSJRv77VLryRnYXvAVtOt655s/edit#heading=h.v83n5y4f1feq)

PR #1730, opened 2024-01-31
Alternative analytics

kdp-cloud:
Adds an alternative way of adding activity trackers by providing a snippet

### DIFF
--- a/app/views/cookies/consent.html.erb
+++ b/app/views/cookies/consent.html.erb
@@ -37,7 +37,7 @@
           <%= link_to t('cookies.buttons.embedding'), cookies_consent_path(allow: 'necessary,embedding'),
                     method: :post, class: 'btn btn-default' %>
           <%= link_to t('cookies.buttons.all'), cookies_consent_path(allow: CookieConsent::OPTIONS.join(',')),
-                    method: :post, class: 'btn btn-default' if (Seek::Config.google_analytics_enabled || Seek::Config.piwik_analytics_enabled || Seek::Config.custom_analytics_snippet_enabled) %>
+                    method: :post, class: 'btn btn-default' if Seek::Config.analytics_enabled %>
         </div>
       </div>
     </div>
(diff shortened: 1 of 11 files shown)


app/views/cookies/consent.html.erb
```diff
@@ -37,9 +37,9 @@
           <%= link_to t('cookies.buttons.embedding'), cookies_consent_path(allow: 'necessary,embedding'),
                     method: :post, class: 'btn btn-default' %>
           <%= link_to t('cookies.buttons.all'), cookies_consent_path(allow: CookieConsent::OPTIONS.join(',')),
-                    method: :post, class: 'btn btn-default' if (Seek::Config.google_analytics_enabled || Seek::Config.piwik_analytics_enabled) %> 
+                    method: :post, class: 'btn btn-default' if (Seek::Config.google_analytics_enabled || Seek::Config.piwik_analytics_enabled || Seek::Config.custom_analytics_snippet_enabled) %>
```
fbacall, line 40: I think maybe consolidate this into a new method in the config class, something like `Seek::Config.analytics_enabled`

and use it in `app/views/layouts/_cookies_banner.html.erb` as well

test/integration/cookie_consent_integration_test.rb
```diff
@@ -164,6 +204,19 @@ class CookieConsentIntegrationTest < ActionDispatch::IntegrationTest
...
 
+  test 'custom analytics code present if only all cookies allowed' do
+    with_config_value(:require_cookie_consent, true) do
+      with_config_value(:custom_analytics_snippet_enabled, true) do
+        post cookies_consent_path, params: { allow: all_options }
+
+        get root_path
+
+        assert CookieConsent.new(cookies).allow_tracking?
+        assert_select '#custom-tracking-script', count: 1
```
fbacall, line 215: In this test I think you should set `custom_analytics_snippet` to something real, with real HTML elements, and then check if you can `assert_select` them - just to make sure they are not being sanitized/escaped.

kdp-cloud: The default value for custom_analytics_snippet is now set to:
```html
<script id="custom-tracking-script"></script>
```

fbacall (approved):

Issue #1743, opened 2024-02-07
Update docker-compose to use MySQL 8


stuzart:
Has some advantages, particularly around handling autoincrement and avoiding reuse of id's if container restarts.

MySQL 8 is already being used for bare-metal deployments with Ubuntu 20 & 22

PR #1744, opened 2024-02-07
update docker compose to use mysql 8.0 #1743

stuzart:
tested updating an existing database, and also starting with a fresh install

the `--log-error-verbosity=1` is to switch of warning logs, and only show errors. There was a deprecation warning about `mysql_native_password` plugin which we may need to address in a later version.
It still shows startup and upgrade logs.

* fix for #1743


kdp-cloud (approved):
I tested it as well and it works for me too.

PR #1749, opened 2024-02-12
Issue 1737 fix isa tags in sample types

kdp-cloud:
- Make sample type attributes editable, based on whether they're inherited.
- Make template attributes editable, based on whether they're inherited.


app/models/template_attribute.rb
```diff
@@ -14,6 +17,18 @@ def controlled_vocab?
     sample_attribute_type.base_type == Seek::Samples::BaseType::CV
   end
 
+  def allow_isa_tag_change?
+    !inherited?
+  end
+
+  def inherited?
+    !parent_attribute_id.nil?
```
stuzart, line 25: would be better to do `parent_attribute.present?` - to handle where the parent id no longer exists, and also looks cleaner

app/views/sample_types/_sample_attribute_form.html.erb
```diff
@@ -22,11 +22,14 @@
...
-<% hide_seek_sample_multi = seek_sample_multi && displaying_single_page? && sample_attribute&.isa_tag.blank? && sample_attribute&.title.downcase.include?('input') %>
-<% isa_tags_options = IsaTag.all.map { |it| [it.title, it.id] } %>
+<% hide_seek_sample_multi = sample_attribute&.input_attribute? && displaying_single_page? %>
+<% level = sample_type.level %>
+<% isa_tags_options = IsaTag.allowed_isa_tags_for_level(level).map { |it| [it.title, it.id] } %>
 <% display_isa_tag ||= false %>
 <% isa_tag_id = sample_attribute&.isa_tag_id %>
 <% isa_tag_title = sample_attribute&.isa_tag&.title %>
+<% template_attribute_id = sample_attribute&.template_attribute_id %>
+<% allow_isa_tag_change = constraints.allow_isa_tag_change?(sample_attribute) %>
```
stuzart, line 32: aren't all these database queries and overheads only necessary if display_isa_tag is true ?

lib/seek/samples/sample_type_editing_constraints.rb
```diff
@@ -75,12 +75,30 @@ def allow_type_change?(attr)
...
+        end
+      end
+
       def refresh_cache
         do_analysis
       end
 
       private
 
+      def inherrited?(attr)
```
stuzart, line 98: misspelling, also could this not be a method and question to the sample attribute? `.inherited_from_template_attribute?`

lib/seek/samples/sample_type_editing_constraints.rb
```diff
@@ -75,12 +75,30 @@ def allow_type_change?(attr)
...
+        attr&.template_attribute_id?
```
stuzart, line 99: would be safer to do `attr&.template_attribute.present?`

stuzart (changes requested):
the inheritance between template attributes as parents to other attribute and sample attributes is a bit confusing. And I'm confused why, as Template also have parents and children. Could the relationship between the attributes not be determined through the parent Template ? Or can attributes from different Templates and SampleTypes have parent/child relationships - which sounds like a nightmare.

There also seems to be danger of parent attributes being deleted, but there children left dangling - not the case for templates as there is a check on can_delete?

kdp-cloud:
Thank you for reviewing and sharing your concerns. I will try to provide a bit of context to why certain choices were made.

Your first question about links between both template attributes and sample attributes comes down to this: I need to know whether a template/sample attribute is inherited from a template or not. We want to prohibit users to edit the ISA tag of a template/sample attribute, when it is inherited. To me it seemed that the best way to achieve this, was by linking the parent attribute and child attribute through a foreign key. If there is a simpler way to achieve the same thing by not making DB changes, please let me know.

Templates have both parents and children because we encourage users to create new templates by starting from existing templates and editing them. This takes away the burden of knowing which ISA tag you need to link to which attribute. The idea of preserving a link between parent and children is meant as a first step towards advanced querying samples through templates. But this feature is not developed yet.

As for deleting parents, it is currently prohibited to delete a template when it has sample types or child templates linked to it. The button should be hidden and there is a clause in the `can_delete?` method to prevent from deleting templates.

In the meantime, I made the changes your requested.

stuzart (approved):

Issue #1784, opened 2024-03-11
flag content blobs as deleted to prevent them being resused


stuzart:
there is a rare case where content blobs may be reused with mysql5. The problem doesn't happen with mysql8 but adding a flag will avoid the problem with older mysql users and overall keep things safe

PR #1785, opened 2024-03-11
flag content blobs to be deleted after asset deleted

stuzart:
set a flag to say the content blob has been deleted, and prevent from showing linked to the asset. This is for cases where the asset_id may be reused, mainly affecting mysql5 but this prevents future surprises.

Current the content blobs aren't deleted, but could be deleted by the regular maintenance job after a set period (perhaps 24hr to make sure there is a chance to back them up). This needs further discussion.

fix for #1784

### DIFF
--- a/app/models/data_file.rb
+++ b/app/models/data_file.rb
@@ -15,7 +15,7 @@ class DataFile < ApplicationRecord
   # allow same titles, but only if these belong to different users
   # validates_uniqueness_of :title, :scope => [ :contributor_id, :contributor_type ], :message => "error - you already have a Data file with such title."
 
-  has_one :content_blob, ->(r) { where('content_blobs.asset_version =?', r.version) }, as: :asset, foreign_key: :asset_id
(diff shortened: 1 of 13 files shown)


app/models/data_file.rb
```diff
@@ -15,7 +15,7 @@ class DataFile < ApplicationRecord
   # allow same titles, but only if these belong to different users
   # validates_uniqueness_of :title, :scope => [ :contributor_id, :contributor_type ], :message => "error - you already have a Data file with such title."
 
-  has_one :content_blob, ->(r) { where('content_blobs.asset_version =?', r.version) }, as: :asset, foreign_key: :asset_id
+  has_one :content_blob, ->(r) { where('content_blobs.asset_version =? AND deleted =?', r.version, false) }, as: :asset, foreign_key: :asset_id
```
fbacall, line 18: I assume this works for existing content blobs with a `NULL` value in the `deleted` column

fbacall (approved):
Just one comment, otherwise looks good

app/models/data_file.rb
stuzart: the migration sets the default to false, which I've always assumed updates previous records. I've checked my local database and this does seem to be the case.

PR #1789, opened 2024-03-15
Extended metadata Write API

whomingbird:


public/api/definitions/_schemas.yml
```diff
@@ -813,11 +828,7 @@ projectMembersList:
     additionalProperties: false
 sampleAttributeMap:
   type: object
-  additionalProperties:
-    anyOf:
-      - type: string
-      - type: array
-      - type: 'null'
```
stuzart, line 820: we need to keep these, so I think you need a separate definition for `extendedMetadataAttributeMap` - ideally with the same restrictions but the addition of the a type for the nested attributes hash, but I'm sure how you would define this so happy with additionalProperties: true for that case, but not here for sample attributes

stuzart (changes requested):

public/api/definitions/_schemas.yml
whomingbird:
```
sampleAttributeMap:
  type: object
  additionalProperties:
      anyOf:
          - type: string
          - type: array
          - type: 'null'
```
i think the indent is not quite right, at least it should looks like 

```
sampleAttributeMap:
  type: object
  additionalProperties: true
  anyOf:
    - type: string
    - type: array
    - type: 'null'
```

stuzart: no that's wrong, the anyOf needs to be part of the additionalProperties as it was before

Issue #1791, opened 2024-03-18
Float sample attribute handlers need to covert, and also expand api tests


stuzart:
Integer and Boolean convert from String to the raw type, but Float doesn't and remains as a String.

The API integration tests for sample needs expanding to include these attribute types, which is likely to then lead to a need to expand the restrictions in the `_schema.yml` for `sampleAttributeMap`

PR #1793, opened 2024-03-19
Convert sample floating point attributes from string to float

stuzart:
in line with boolean and integer conversions.
Expanded max_sample and max_sample_type to include boolean, float and integer and expand api test coverage, requiring an update of the api schema to support these types for the sample attribute map

* fix for #1791

### DIFF
--- a/public/api/definitions/_schemas.yml
+++ b/public/api/definitions/_schemas.yml
@@ -833,6 +833,8 @@ sampleAttributeMap:
       - type: string
       - type: array
       - type: 'null'
+      - type: number
+      - type: boolean
 extendedMetadataAttributeMap:
   type: object
   additionalProperties: true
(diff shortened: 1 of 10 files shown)


public/api/definitions/_schemas.yml
whomingbird, line 1: should also add -type: integer ?

whomingbird: Ok, if you convert all `integer` to `float`, `-type: integer` is not needed.

whomingbird (approved):

public/api/definitions/_schemas.yml
stuzart: integers aren't being converted to floats. But the json schema type 'number' also accepts integers - https://json-schema.org/understanding-json-schema/reference/numeric#number

PR #1804, opened 2024-03-25
Options to reduce the amount of logging with docker

stuzart:
a couple of environment variables added:

SEEK_LOG_LEVEL to set the rails log levels
QUIET_SUPERCRONIC - if set will rub supercronic is quiet mode and avoid the regular ticking over of log messages

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -50,7 +50,7 @@ Rails.application.configure do
 
   # Include generic and useful information about system operation, but avoid logging too much
   # information to avoid inadvertent exposure of personally identifiable information (PII).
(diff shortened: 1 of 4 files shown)


fbacall (commented):
In recent versions of Rails they use `RAILS_LOG_LEVEL`. Maybe worth using the same variable name for compatibility?
https://github.com/rails/rails/blob/246b3b609b130072ea009b3f2b74e603e52ac850/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L73

fbacall (approved):

Issue #1837, opened 2024-04-18
Show the batch permission changes, and publish button on the My Items page


stuzart:
The are currently on the profile page, which isn't that intuitive. They could also be shown on the My Items page

jjkoehorst:
Or when you change the permission of a single item? Have a button that shows there is this option for bulk edit?

PR #1845, opened 2024-04-18
Batch permission on my items 

stuzart:
Show the batch permission and publish buttons on the My Items page

* fix for #1837 

also a little refactoring, removing the `mine?` helper method

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -273,22 +273,6 @@ module ApplicationHelper
     contributor_name_link.html_safe
   end
 
-  # this helper is to be extended to include many more types of objects that can belong to the
-  # user - for example, SOPs and others
-  def mine?(thing)
-    return false if thing.nil?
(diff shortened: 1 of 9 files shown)


kdp-cloud (approved):
It looks OK, so I'll approve. But I'm just wondering why the `mine?` function had to go.
Was it redundant? Why was it introduced in the first place?
It looks like someone wanted to catch the case where an item belongs to a user but does not have editing permissions.

stuzart:
> It looks OK, so I'll approve. But I'm just wondering why the `mine?` function had to go. Was it redundant? Why was it introduced in the first place? It looks like someone wanted to catch the case where an item belongs to a user but does not have editing permissions.

it was a very old method that didn't do how it was described. It always returned false unless the object was a Person, in which case it did the same as Person.me? . There were cases where it was used and would always return false, so .can_edit? was being called instead.

PR #1857, opened 2024-04-24
Sanitize objects input dynamic table

kdp-cloud:
Fix for XSS vulnerability in dynamic table in multiple places.

### DIFF
--- a/app/assets/javascripts/single_page/dynamic_table.js.erb
+++ b/app/assets/javascripts/single_page/dynamic_table.js.erb
@@ -578,7 +578,7 @@ function inputObjectsInput(column, data, options, url){
     }
   }).join("");
   if (options.readonly) {
-    return data.map((e) => `<span title="ID: ${e.id}" class="badge">${e.title}</span>`).join(" ");
+    return data.map((e) => `<span title="ID: ${e.id}" class="badge">${sanitizeData(e.title)}</span>`).join(" ");
   } else {
     const typeaheadTemplate = 'typeahead/single_pages_samples'
     const objectInputName = data.map((e) => e.id).join('-') + '-' + crypto.randomUUID();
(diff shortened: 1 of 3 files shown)


kdp-cloud:
HTML tags in sample attribute values are escaped now.
image

app/helpers/dynamic_table_helper.rb
```diff
@@ -159,7 +159,9 @@ def dt_cumulative_cols(sample_types)
         attribute = { title: a.title, name: s.id.to_s, required: a.required, description: a.description,
                       is_title: a.is_title }
         condition = a.sample_attribute_type.base_type == Seek::Samples::BaseType::SEEK_SAMPLE_MULTI
+        is_cv_list = a.sample_attribute_type.base_type == Seek::Samples::BaseType::CV_LIST
```
stuzart, line 162: you can use `sample_attribute_type.seek_cv_list?`

app/assets/javascripts/single_page/dynamic_table.js.erb
```diff
@@ -582,16 +610,16 @@ function inputObjectsInput(column, data, options, url){
...
-    const titleText = '';
-    const allowNewItems = column.cv_allows_free_text;
-    setTimeout(ObjectsInput.init);
+  const existingOptions = data.map((e) => `<option selected="selected" value="${sanitizeData(e)}">${sanitizeData(e)}</option>`);
+    if (options.readonly) {
+        return data.map((e) => `<span title="ID: ${sanitizeData(e)}" class="badge">${sanitizeData(e)}</span>`).join(" ");
+    } else {
+        const typeaheadTemplate = 'typeahead/controlled_vocab_term';
+        const objectInputName = data.map((e) => sanitizeData(e)).join('-') + '-' + crypto.randomUUID();
+        const extraClass = '';
```
stuzart, line 619: could `sanitizeData(e)` be called once at the start and assigned to a variable, instead of being called repeatedly?

stuzart (changes requested):
minor changes

app/assets/javascripts/single_page/dynamic_table.js.erb
stuzart: looking at it again, I can see that e may be different in each case, so ignore this

stuzart (approved):

PR #1872, opened 2024-05-08
Issue 1783 fix add new button

### LINKED ISSUE #1783: Remove entries from "Add new" in DataHub
In DataHub, we do not (yet) wish to make use of some entities such as strain and model. These entries have been removed (disabled, hidden) from Browse and Create, but they are still accessible via "Add new".

Fix: remove not needed entries from "add new", for consistency with Browse and Create options in DataHub

Image

kdp-cloud:
The Add new drop-down menu hides 'Add new Strain' when `organisms_enabled` is set to false. Fixes #1783


stuzart (changes requested):
A new config setting probably isn't appropriate, since strains is dependant on organisms.
Just add a `strains_enabled` method to the Seek::Config that returns `organism_enabled` .

There are other cases here: https://github.com/seek4science/seek/blob/main/lib/seek/config.rb#L323 

e.g how `templates_enabled `is dependant on `isa_json_complaint_enabled`

stuzart (approved):

PR #1884, opened 2024-05-16
Bugfixes Sample Experiment Templates

### LINKED ISSUE #1757: Remove GUI to create ISA-compliant Templates from scratch
Remove GUI to create  ISA-compliant Templates from scratch.
User of GUI can only create  ISA-compliant Templates starting from existing ISA-compliant Templates.

### LINKED ISSUE #1758: Reduce amounts of checks when creating a template
Currently when creating a new template, based on an existing template, the inherited attributes are completely locked because it needs to be locked when creating a Sample Type. However, creating a new template, based on an existing one, does not mean it has to comply to the parent template but is regarded as a service to the user.

__To do__:
The javascript that deals with locking elements should behave differently when creating a new template.
- All attributes should be removable
- Only the ISA tag should be locked (as it was in the past)

### LINKED ISSUE #1880: Template creation if non-administrator
A User cannot create templates unless he administers at least 1 project. Any user should be able to create a template, regardless if the user is administrating anything.

kdp-cloud:
- Makes sure users cannot create new templates from scratch. Closes #1757 
- Template behavior is now different from Sample type behavior. Some limitations of the Sample Type are not applied on templates anymore:
  - Template attributes can be removed when inherrited from another template
  - All fields (except ISA tag) are editable again for inherrited attributes
  - Closes #1758 
- Users can only create templates and link them to projects they administer instead of all projects they are member of. Closes #1880

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -54,6 +54,7 @@ class ApplicationController < ActionController::Base
   def current_person
     current_user.try(:person)
   end
+  helper_method :current_person
 
   def partially_registered?
     redirect_to register_people_path if current_user && !current_user.registration_complete?
--- a/app/assets/javascripts/templates.js
+++ b/app/assets/javascripts/templates.js
@@ -293,9 +293,12 @@ const applyTemplate = () => {
   const template_id_tag = $j(`#isa_study${suffix}template_parent_id`);
   if (template_id_tag) $j(template_id_tag).val(id);
 
+  // Removes the hidden from the new attribute button
+  $j(`${attribute_table} ${addAttributeRow}`).find('#add-attribute').removeClass("hidden");
+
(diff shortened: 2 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -54,6 +54,7 @@ def with_current_user
   def current_person
     current_user.try(:person)
   end
+  helper_method :current_person
```
stuzart, line 57: this should be moved to the top of the file, next to `helper :all`

stuzart (changes requested):
minor change

app/controllers/application_controller.rb
kdp-cloud: I grouped all helper methods together.

stuzart (approved):

PR #1888, opened 2024-05-21
Don't show non-editable assays when extracting samples

### LINKED ISSUE #1887: All assays shown to user when extracting samples, not just those with edit permissions
When extracting samples from a data file, all of the assays associated to the data file are shown to the user when reviewing the extracted samples, not just those with edit permissions.

Note: even if a user selects all visible assays, only those which the user is permitted to edit will be associated with the extracted samples, so this is purely a visual issue on the confirm_extraction page.

AndrewWood94:
- Add authorised_assays to datafiles_helper 
- Only show users authorised_assays when extracting samples

Resolves #1887

### DIFF
--- a/app/helpers/data_files_helper.rb
+++ b/app/helpers/data_files_helper.rb
@@ -3,6 +3,10 @@ module DataFilesHelper
     authorised_assets(DataFile, projects)
   end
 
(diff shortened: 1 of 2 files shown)


stuzart (approved):
thanks.

PR #1898, opened 2024-05-27
Fix batch sharing permissions in single page

### LINKED ISSUE #1892: Batch sharing permission in `Experiment view` doesn't work anymore
When selecting samples and clicking batch sharing permission, the modal form pops up and shows the checked samples.
image

It does not let the user add additional permissions (user, project, institution, ...) to the permissions. After clicking the 'add' button, the modal forms simply disappears.
image

kdp-cloud:
- Modified the modal form that allows multiple opened modal forms.
- Fixes #1892

### DIFF
--- a/app/helpers/bootstrap_helper.rb
+++ b/app/helpers/bootstrap_helper.rb
@@ -161,10 +161,17 @@ module BootstrapHelper
     end
   end
 
-  def modal_header(title, _options = {})
+  def modal_header(title, options = {})
+    # If no_close_btn is set to true, the modal will not have a close button.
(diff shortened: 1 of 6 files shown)


stuzart (commented):
As mentioned in the meeting, I think you can solve the model problem simply by adding these lines to your base view to prevent the events being propogated:

https://github.com/seek4science/seek/blob/main/app/views/data_files/_provide_metadata.html.erb#L64-L68

if you are also showing the sharing dialog as a modal, you may need an additional one to cover that dialogues data-disimss event

stuzart (approved):

PR #1925, opened 2024-06-19
Unzip datafiles feature

AndrewWood94:
Adds ability to unzip datafiles and save contents as new datafiles, which can be linked to any assay the zipped folder is linked to.

Notes:
- A way to link multiple datafiles to an assay simultaneously would be useful if users forget to link in the unzip process
- Nothing is implemented to handle new datafile versions not matching the original zip contents, this could be added in a similar manner to updating extracted samples
- Unzipped datafiles do not have a creator, similar to extracted samples
- Currently no tests have been written for this feature

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -50,6 +50,7 @@
 //= require DataTables-1.11.3/buttons.html5
 //= require DataTables-1.11.3/autoFill.bootstrap
 //= require samples
+//= require datafiles
 //= require jstree
 //= require clipboard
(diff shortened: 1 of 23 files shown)


stuzart (approved):
code is good, and gave it a run through and test and it works quite nicely. thanks.

I will merge this and add some tests on a separate branch (since I can't push to this one). 

If you agree, I also thought I could make the creators default to the same as the original zip file. I was also thinking of showing the origin zip file as part of the metadata on the data file show page.

Issue #1950, opened 2024-07-09
FAIR data station enabled configuration


stuzart:
FAIR data station enabled configuration, which when disabled hides the option to start an import (from the project action menu) and also block the actions to start and submit the import.
When disabled, Observation Units will also be disabled, hidden and blocked.

Issue #1949, opened 2024-07-09
Import FAIR data station TTL metadata


stuzart:
For a given project, be able to import the TTL metadata file, that has been created when validating the FAIR data station template.
This will construct and import the Investigation->Study->ObservationUnit->Sample->Assay model with the relevant core metadata. Where there is additional metadata it should attempt to find and use a relevant Extended Metadata definition.
When being imported, it will be associated with the Project and have the ability to set the sharing permissions

Issue #1948, opened 2024-07-09
Allow sheet name when getting csv, and better error reporting if invalid sheet number or name


stuzart:
When exporting an Excel data file as CSV, allow the sheet name to be specified by name as well as by index. Also improve the response code and error reporting when a sheet is either out of index or it's name is not recognised

Issue #1947, opened 2024-07-09
Observation Unit read and write API


stuzart:
JSONAPI for Observation Unit, to fully support read and write, including full test coverage. Schema and documentation will come later

PR #1951, opened 2024-07-09
FAIR data station support, and related work, as part of the BioIndustry Prototype

stuzart:
Support for importing metadata from the FAIR Data Station, to construct entries within SEEK. The metadata is a Turtle (TTL) file generating via validiting a metadata template through the FAIR Data Station.
Some parts are still a work in progress, but can be disabled through `Seek::Config.fair_data_station_enabled` and is disabled by default.

Addresses the following issues:

* #1949
* #1946 
* #1947 
* #1945 
* #1944 
* #1948 
* #1950


app/models/investigation.rb
```diff
@@ -53,4 +53,22 @@ def positioned_studies
   def self.user_creatable?
     Seek::Config.investigations_enabled
   end
+
+  # utility method for cleaning up durng testing, not to be used in production code
+  def deep_destroy!
```
whomingbird, line 58: I've tried using `deep_destroy!` during the review, but I noticed that the data files and samples are not being destroyed after running this method.

public/api/definitions/_schemas.yml
```diff
@@ -837,6 +837,7 @@ sampleAttributeMap:
       - type: 'null'
       - type: number
       - type: boolean
+      - type: hash
```
fbacall, line 840: should be `type: object` I think

fbacall: or maybe even remove this specification since it includes every possible type.

See also: https://github.com/seek4science/seek/blob/main/public/api/definitions/_schemas.yml#L840-L842

maybe could re-use that?

app/models/investigation.rb
stuzart: Updated and fixed. Also extended to delete samples and data files associated with observation units, which lead to an issue deleting `ObservationUnitAsset` which has also been fixed

test/fixtures/files/fairdatastation/demo.ttl
```diff
@@ -0,0 +1,877 @@
...
+        jerm:hasPart        <http://fairbydesign.nl/ontology/inv_INV_DRP007092/stu_DRP007092> ;
+        schema:contributor  <http://fairbydesign.nl/ontology/inv_INV_DRP007092/person/343732a9426f424e535356c966739de9> ;
+        schema:description  "Exploration of HIV-1 infected individuals in Ghana" ;
+        schema:identifier   "INV_DRP007092" ;
+        schema:title        "HIV-1 infected individuals in Ghana" .
+
+<http://fairbydesign.nl/ontology/inv_INV_DRP007092/person/343732a9426f424e535356c966739de9>
+        rdf:type           schema:Person ;
+        schema:department  "AIDS Research Center" ;
+        schema:email       <mailto:pkofieparbie@nih.go.jp> ;
```
fbacall, line 39: Maybe worth replacing with a dummy email address

test/functional/observation_units_controller_test.rb
```diff
@@ -0,0 +1,133 @@
...
+  test 'show' do
+    unit = FactoryBot.create(:max_observation_unit)
+    get :show, params: { id: unit.id }
+    assert_response :success
+  end
+
+  test 'index' do
+    unit = FactoryBot.create(:max_observation_unit)
+    get :index
+    assert_response :success
```
fbacall, line 23: Also check the obs unit is in there.

test/functional/projects_controller_test.rb
```diff
@@ -5042,6 +5042,101 @@ def test_admin_can_edit
...
+    with_config_value(:fair_data_station_enabled, false) do
+      get :show, params: { id: project.id }
+      assert_response :success
+      assert_select '#item-admin-menu' do
+        assert_select 'li a[href=?]', import_from_fairdata_station_project_path(project), text:/Import from FAIRData Station/, count: 0
+      end
+
+      get :import_from_fairdata_station, params: { id: project.id }
+      assert_redirected_to :root
+      refute_nil flash[:error]
```
fbacall, line 5077: Check error message because it could just be because there is no user logged in

fbacall (changes requested):

app/views/observation_units/_resource_list_item.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= list_item_description resource.description %>
+
+<div class="row">
+  <div class="col-sm-6">
+
+  </div>
+</div>
```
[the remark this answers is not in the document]
stuzart: I'd left it as a placeholder until it's decided what is shown, but I've removed it for now and can always be put back

lib/seek/config.rb
```diff
@@ -340,6 +340,10 @@ def strains_enabled
       organisms_enabled
     end
 
+    def observation_units_enabled
+      fair_data_station_enabled
+    end
```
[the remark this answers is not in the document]
stuzart: Not currently or for the forseable future. They were added to support the FAIR data station model. You can't actually currently create them without importing from FAIR data station.
If there is a need in the future to use them without FDS then the config can be easily flipped over.

test/fixtures/files/fairdatastation/demo.ttl
stuzart: it's a public dataset I checked could be shared, but I've replaced it anyway.

config/version.yml
```diff
@@ -9,4 +9,4 @@
 
 major:        1
 minor:        16
-patch:        0-main 
+patch:        0-bioindustry
```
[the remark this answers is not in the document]
stuzart: yeah this will be handled when merged

lib/seek/fair_data_station/base.rb
```diff
@@ -0,0 +1,114 @@
...
+        annotations.reject do |annotation|
+          core_annotations.include?(annotation[0])
+        end
+      end
+
+      def find_annotation_value(property)
+        annotations.detect do |ann|
+          ann[0] == property
+        end&.[](1)
+      end
```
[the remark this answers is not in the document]
stuzart: cache the annotations for now. There might be a case for some optimisations but not until a later stage when we know what kind of data we get and whether any optimisations are needed. Currently we are expecting many resources with only a small number of properties, to just iterating may be quicker than a sparql query over a large graph each time. If he have only a few resources with many many properties that might change.

public/api/definitions/_schemas.yml
stuzart: consolidated into a single `attributeMap` definition

app/controllers/projects_controller.rb
```diff
@@ -195,6 +199,19 @@ def request_join
...
+    policy.set_attributes_with_sharing(policy_params)
+    datadata_inv = Seek::FairDataStation::Reader.instance.parse_graph(path)
+    @investigation = Seek::FairDataStation::Writer.instance.construct_isa(datadata_inv.first, current_person, [@project], policy)
+    @investigation.save!
+
+    respond_to do |format|
+      format.html { redirect_to(@investigation) }
+    end
+  end
+
```
whomingbird, line 214: if the TTL file contains syntax errors, how to handle it?

whomingbird:
I've noticed that users can import the same TTL file multiple times, resulting in duplicate ISAs within the project. To address this, we should consider the following enhancements for future updates:

1. Allow users to delete the entire ISA tree along with its associated content, in case they mistakenly import the data more than once. (`deep_destroy! ` )

2. Implement the import process as a delayed job to avoid prolonged response times, ensuring users are aware of the import status. This will also help prevent repeated, mistaken imports of the same data.

app/controllers/projects_controller.rb
stuzart: currently it will throw an error, but that is fine for now. The TTL is being generated by FDS so unlikely to happen

stuzart:
> I've noticed that users can import the same TTL file multiple times, resulting in duplicate ISAs within the project. To address this, we should consider the following enhancements for future updates:
> 
> 1. Allow users to delete the entire ISA tree along with its associated content, in case they mistakenly import the data more than once. (`deep_destroy! ` )
> 2. Implement the import process as a delayed job to avoid prolonged response times, ensuring users are aware of the import status. This will also help prevent repeated, mistaken imports of the same data.

Yes, as the the PR description description states this is still a work in progress and is disabled by default. Further work will continue on the main branch. The aim here is, having reached a milestone, to avoid having a long running branch and an even bigger ever growing PR. Futher tasks are sketched out here - https://github.com/seek4science/seek/projects/19 including avoiding duplication.

whomingbird:
`Observation units `are not listed in the "publish related items" 
e.g. http://localhost:3000/investigations/7/publish_related_items
could add it as the future work.

public/api/examples/studiesResponse.json
whomingbird, line 1: Currently the type observation unit is not listed in the the Study read API.

whomingbird (approved):

fbacall (approved):

PR #1953, opened 2024-07-10
Issue 1199 add sharing permisions to sample types

kdp-cloud:
- Users are now able to set policies with permissions.
- The current rules for viewing, editing and deleting should still remain the same.
- There is also a upgrade job that adds policies to the existing sample types.


lib/tasks/seek_upgrades.rake
```diff
@@ -235,6 +236,40 @@ namespace :seek do
...
+  task(add_policies_to_existing_sample_types: [:environment]) do
+    puts '... Adding policies to existing sample types'
+    counter = 0
+    disable_authorization_checks do
+      SampleType.includes(:projects, :assays, :studies).where(policy_id: nil).each do |st|
+        if st.is_isa_json_compliant?
+          st.update_column(:policy_id, st.assays.first.policy_id) if st.assays.any?
+          st.update_column(:policy_id, st.studies.first.policy_id) if st.studies.any?
+        else
+          project = st.projects.first
```
stuzart, line 248: what about multiple projects ?

lib/tasks/seek_upgrades.rake
```diff
@@ -235,6 +236,40 @@ namespace :seek do
...
+          if project.use_default_policy
+            st.update_column(:policy_id, project.default_policy_id)
```
stuzart, line 250: this will create a reference to the default policy, which will then also be changed if the ST sharing permissions are changed. I'm also not sure it's right

lib/tasks/seek_upgrades.rake
```diff
@@ -235,6 +236,40 @@ namespace :seek do
...
+          else
+            policy = Policy.new
+            policy.name = 'default policy'
+            policy.access_type = Policy::NO_ACCESS
+            policy.permissions << Permission.create(contributor_type: Permission::PERSON, contributor: st.contributor,
```
stuzart, line 255: contributor will already have manage permissions by default

lib/tasks/seek_upgrades.rake
```diff
@@ -235,6 +236,40 @@ namespace :seek do
...
+                                                    access_type: Policy::MANAGING)
+            project.people.each do |person|
```
stuzart, line 257: should just add the permission for the project, not each person.

lib/tasks/seek_upgrades.rake
```diff
@@ -235,6 +236,40 @@ namespace :seek do
     end
   end
 
+  task(add_policies_to_existing_sample_types: [:environment]) do
```
stuzart, line 239: I think generally this needs a rethink and currently doesn't reflect the original permissions.

Would need checking more closely, but generally:
* make visible if linked to public samples, otherwise hidden
* add visible permissions for each project
* give manage rights to the project administrators

stuzart (changes requested):
I think the main problem is the mixing of explicit permissions then being over-ruled by the old implicit permissions

I've not checked the tests as these may be affected by changes

lib/seek/acts_as_asset/content_blobs.rb
```diff
@@ -5,7 +5,7 @@ module ContentBlobs
       module ClassMethods
         extend ActiveSupport::Concern
         included do
-          after_destroy :mark_deleted_content_blobs
+          after_destroy :mark_deleted_content_blobs, unless: -> { self.class == SampleType }
```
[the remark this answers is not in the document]
kdp-cloud: Because this test fails:
https://github.com/seek4science/seek/blob/8463e2774e021c3d4c675c1bf3c68611644bbd31/test/unit/sample_type_test.rb#L517-L536

```
Error:
SampleTypeTest#test_dependant_destroy_content_blob:
ActiveRecord::ActiveRecordError: cannot update a destroyed record
    lib/seek/acts_as_asset/content_blobs.rb:57:in `block in mark_deleted_content_blobs'
    lib/seek/acts_as_asset/content_blobs.rb:56:in `each'
    lib/seek/acts_as_asset/content_blobs.rb:56:in `mark_deleted_content_blobs'
    test/unit/sample_type_test.rb:459:in `block (5 levels) in <class:SampleTypeTest>'
    test/unit/sample_type_test.rb:458:in `block (4 levels) in <class:SampleTypeTest>'
    test/unit/sample_type_test.rb:457:in `block (3 levels) in <class:SampleTypeTest>'
    app/models/user.rb:271:in `with_current_user'
    test/unit/sample_type_test.rb:449:in `block (2 levels) in <class:SampleTypeTest>'
    test/test_helper.rb:77:in `with_config_value'
    test/unit/sample_type_test.rb:448:in `block in <class:SampleTypeTest>'
```

I assume it has something to do with this line:
https://github.com/seek4science/seek/blob/8463e2774e021c3d4c675c1bf3c68611644bbd31/lib/seek/templates/sample_type_template_concerns.rb#L10

I didn't want to mess too much with it, so I skipped sample types to be safe.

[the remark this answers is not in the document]
stuzart: yes, that would be why, I think would be better to remove the `depedent: destroy` and keep the other behaviour consistent

test/functional/samples_controller_test.rb
```diff
@@ -793,56 +793,56 @@ def rdf_test_object
...
+  #
+  #   get :show, params: { id:sample2.id }
+  #   assert_response :success
+  #
+  #   # no referring sample required
+  #   assert_select 'a[href=?]',sample_type_path(sample_type2),text:/#{sample_type2.title}/
+  #
+  # end
+
+  # test 'referring sample id is added to sample type links in list items' do
```
stuzart, line 824: same as previous comment

app/controllers/sample_types_controller.rb
```diff
@@ -2,15 +2,17 @@ class SampleTypesController < ApplicationController
   respond_to :html, :json
   include Seek::UploadHandling::DataUpload
   include Seek::IndexPager
+  include Seek::AssetsCommon
 
   before_action :samples_enabled?
-  before_action :find_sample_type, only: [:show, :edit, :update, :destroy, :template_details, :batch_upload]
+  before_action :find_sample_type, only: %i[show edit update destroy template_details batch_upload]
```
stuzart, line 8: because of the order actions are run, this will likely be overridden by find_and_authorize_requested_item so it possibly no longer needed.

But needs checking because the logic is slightly different wrt isa_json_compliant sample types

stuzart (changes requested):

test/functional/samples_controller_test.rb
```diff
@@ -793,56 +793,56 @@ def rdf_test_object
...
-
-    get :index
-
-    assert_select 'a[href=?]',sample_type_path(sample_type,referring_sample_id:sample.id),text:/#{sample_type.title}/
-
-    # no referring sample required, ST is already visible
-    assert_select 'a[href=?]',sample_type_path(sample_type2),text:/#{sample_type2.title}/
-
-  end
+  # test 'referring sample id is added to sample type link, if necessary' do
```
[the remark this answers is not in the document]
stuzart: sample_type_link method in samples_helper might need updating to hide the sample type title and link if not visible

app/controllers/sample_types_controller.rb
kdp-cloud: I double checked it and it doesn't get skipped. In fact, it makes some tests fail if I remove this method.

test/functional/samples_controller_test.rb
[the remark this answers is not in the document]
kdp-cloud: Adding `person` as contributor to the Sample Type makes the tests pass. I uncommented it again.

stuzart (approved):

PR #1954, opened 2024-07-10
Log workflow and model executions

fbacall:
Records an `ActivityLog` when a workflow or JWS-compatible model is run and displays a count along with view and downloads on the right-hand side of the show page


whomingbird (commented):
When running the JWS simulation once, the "run" log increases by 2 because both URL requests add an entry to the run log.

http://0.0.0.0:3000/models/1/simulate?version=1
http://0.0.0.0:3000/models/1/simulate?version=1&constraint_based=0

whomingbird (changes requested):
The "run" logs increase correctly for workflows in execution, but not for model simulations.

whomingbird (approved):

PR #1989, opened 2024-09-06
Issue 1977 query by est fix reg samples

### LINKED ISSUE #1977: Querying by EST - registered (input) samples does not work
Querying by Experiment Sample Template does not return any samples when a registered (input) sample is used as a query.

Browsing samples by Experiment Sample Template is possible by querying according to different attributes, including String, Output sample, CV, … However, whenever an input sample attribute is selected, the search returns zero samples. 

Ps: Searching with wildcards within ‘Query by Experiment Sample Template’ also does not work. It does work for the ‘normal’ search function ([Issue #1827](https://github.com/seek4science/seek/issues/1827))

Screenshots
Query for samples using the assay sample template returns 176 samples, showing one sample using input sample name “ss_TMDsel_0001”:
image

Filtering samples using the assay sample template using input “ss_TMDsel_0001” (0 samples):
image

Filtering samples using the assay EST using advanced filtering (Study sample; output name = “ss_TMDsel_0001” (1 sample):
image

Filtering samples using the assay EST using advanced filtering (Study sample; input name = “s_TMDsel_0001” (0 samples):
image

Filtering samples using the assay EST using advanced filtering (Study source; output name = “s_TMDsel_0001” (1 sample):
image

Details can also be found in the following [Google Document](https://docs.google.com/document/d/1507RWawIHXMHo9YcJd38aXWJRf2YtGLxWRqDW8Ri47E/edit)

kdp-cloud:
Fixes #1977 

- Querying samples by attributes that are not of type string, like SEEK samples and CV lists, is now possible.
- The query matches are now not case-sensitive anymore

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -244,21 +244,21 @@ class SamplesController < ApplicationController
     end
 
     if params[:input_template_id].present? # linked
-      title =
-        TemplateAttribute.find(params[:input_attribute_id]).title if params[:input_attribute_id].present?
-      @result = find_samples(@result, :linked_samples,
-        { attribute_id: params[:input_attribute_id],
+      input_template_attribute =
+        TemplateAttribute.find(params[:input_attribute_id])
+      @result = filter_linked_samples(@result, :linked_samples,
+                                      { attribute_id: params[:input_attribute_id],
           attribute_value: params[:input_attribute_value],
-          template_id: params[:input_template_id] }, title)
+          template_id: params[:input_template_id] }, input_template_attribute&.title)
     end
 
     if params[:output_template_id].present? # linking
-      title =
-        TemplateAttribute.find(params[:output_attribute_id]).title if params[:output_attribute_id].present?
-      @result = find_samples(@result, :linking_samples,
-        { attribute_id: params[:output_attribute_id],
+      output_template_attribute =
+        TemplateAttribute.find(params[:output_attribute_id])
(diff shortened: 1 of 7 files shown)


app/controllers/samples_controller.rb
```diff
@@ -318,16 +332,34 @@ def find_index_assets
...
+  # @param options [Hash] the options for filtering
+  # @option options [Integer] :template_id the ID of the template to filter by
+  # @option options [String] :attribute_value the value of the attribute to filter by
+  # @option options [Integer] :attribute_id the ID of the attribute to filter by
+  # @param template_attribute_title [String] the title of the template attribute to filter by
+  # @return [Array<Sample>] the filtered list of samples
+  def filter_linked_samples(samples, link, options, template_attribute)
+    template_attribute_title = template_attribute&.title
     samples.select do |s|
       s.send(link).any? do |x|
```
stuzart, line 348: i think this is potentially dangerous, maybe not now but could be due to an error in the future. Just needs to check the `link` is in a list of allowed method calls, or have another way of specifying which method is called.

stuzart (changes requested):

app/controllers/samples_controller.rb
kdp-cloud: Indeed! makes sense.
Now it will raise an error if `link` is not either `:linked_samples` or `:linking_samples`.

stuzart (approved):

Issue #1991, opened 2024-09-10
Enhance UI to Better Display Control Vocabulary Attribute Values with IRI References


whomingbird:
![image](https://github.com/user-attachments/assets/e476c43a-b4d6-4c4c-98fc-52ae755bedb7)

The current controlled vocabulary (CV) values associated with ontology terms are displayed as plain text. We would like to enhance these by converting them into hyperlinks, pointing to their respective ontology term URLs. This will improve usability by allowing users to easily access detailed information about the ontology terms.

from Biomedical science (http://edamontology.org/topic_3344)  to [Biomedical science](http://edamontology.org/topic_3344)

stuzart:
bear in mind that the ontology identifiers may not always be resolvable URLs, or may just resolve to an OWL xml file. Might be better to make the hyperlinks point to a reference in the OLS

stuzart:
> Is it currently supported in SEEK to define an `iri` that points to an OWL XML file or non-resolvable URLs? [ reference](https://github.com/seek4science/seek/blob/seek-1.15/config/default_data/data-annotations-controlled-vocab.json)

yes, the iri dependes on the the ontology and what was chosen for the term id iri. An IRI isn't a URL, `fred:blogs` is a valid IRI, as is `chebi:1234` or `taxon:1234`. So you can't rely on the ontology using resolvable urls.

http://edamontology.org/topic_3344 only resolves becomes somebody has taken the effort to setup a redirect to bioportal, but they didn't have to do so.

stuzart:
here is an example of a term that resolves to an owl.xml file: https://www.ebi.ac.uk/ols4/ontologies/enm/classes/http%253A%252F%252Fwww.bioassayontology.org%252Fbao%2523BAO_0000015

( http://www.bioassayontology.org/bao#BAO_0000015 )

stuzart:
> here is an example of a term that resolves to an owl.xml file: https://www.ebi.ac.uk/ols4/ontologies/enm/classes/http%253A%252F%252Fwww.bioassayontology.org%252Fbao%2523BAO_0000015
> 
> ( http://www.bioassayontology.org/bao#BAO_0000015 )

or a better link to the ontology itself, where all terms to: https://www.ebi.ac.uk/ols4/ontologies/bao

PR #1992, opened 2024-09-11
Update CV Terms to Hyperlinked Ontology References for Extended Metadata…

whomingbird:
#1991

### DIFF
--- a/app/helpers/samples_helper.rb
+++ b/app/helpers/samples_helper.rb
@@ -1,4 +1,6 @@
 module SamplesHelper
+  include Seek::UrlValidation
+
   def sample_form_field_for_attribute(attribute, resource)
     element_class = "sample_attribute_#{attribute.sample_attribute_type.base_type.downcase}"
     element_name = "sample[data][#{attribute.title}]"
@@ -143,7 +145,9 @@ module SamplesHelper
       when Seek::Samples::BaseType::CV
         seek_cv_attribute_display(value, attribute)
       when Seek::Samples::BaseType::CV_LIST
-        value.each{|v| seek_cv_attribute_display(v, attribute) }.join(', ')
+        value.map do |v|
+          seek_cv_attribute_display(v, attribute)
+        end.join(', ').html_safe
       when Seek::Samples::BaseType::LINKED_EXTENDED_METADATA
         linked_extended_metadata_attribute_display(value, attribute)
       when Seek::Samples::BaseType::LINKED_EXTENDED_METADATA_MULTI
@@ -172,11 +176,11 @@ module SamplesHelper
 
   def seek_cv_attribute_display(value, attribute)
     term = attribute.sample_controlled_vocab.sample_controlled_vocab_terms.where(label:value).last
-    content = value
-    if term && term.iri.present?
-      content << " (#{term.iri}) "
+    if term && term.iri.present? && valid_url?(term.iri)
+      link_to(term.label, term.iri, target: :_blank)
+    else
+      term.label
     end
-    content
   end
 
   def linked_extended_metadata_attribute_display(value, attribute)


app/assets/javascripts/sample_types.js
```diff
@@ -96,6 +96,11 @@ var SampleTypes = {
         else {
             seek_sample_element.hide();
         }
+    },
+
+    expandIRI: function () {
+        var termIRI = $j(this).next('.term-iri');
+        termIRI.toggleClass('visible');
```
stuzart, line 103: might be better to use `.show()` from jquery, as this is what we use elsewhere and is a bit clearer

app/helpers/samples_helper.rb
```diff
@@ -143,7 +143,9 @@ def display_attribute_value(value, attribute, options = {})
       when Seek::Samples::BaseType::CV
         seek_cv_attribute_display(value, attribute)
```
stuzart, line 144: have you tested with the single cv case? I wonder whether this may also now need the `.html_safe`

app/helpers/samples_helper.rb
```diff
@@ -171,12 +173,15 @@ def select_cv_source_ontology(sample_controlled_vocab)
   end
 
   def seek_cv_attribute_display(value, attribute)
-    term = attribute.sample_controlled_vocab.sample_controlled_vocab_terms.where(label:value).last
-    content = value
+    term = attribute.sample_controlled_vocab.sample_controlled_vocab_terms.where(label: value).last
     if term && term.iri.present?
-      content << " (#{term.iri}) "
+      iri_content = term.iri.match?(/^http/) ? link_to(term.iri, term.iri, target: '_blank') : term.iri
```
stuzart, line 178: `/^https?:\/\//` would be a bit safer as the regular expression

app/views/extended_metadata/_extended_metadata_attribute_values.html.erb
```diff
@@ -35,5 +35,10 @@
         window.collapseAll = function() {
             panelcollaps.removeClass('in');
         };
+
+        $j('.term-label').on('click', SampleTypes.expandIRI);
```
stuzart, line 39: this can be removed if the next comment is addressed ....

app/views/samples/_attribute_values.html.erb
```diff
@@ -4,3 +4,9 @@
       <%= display_attribute(sample, attribute, link: true) %>
     </p>
 <% end %>
+
+<script>
+    $j(document).ready(function() {
+        $j('.term-label').on('click', SampleTypes.expandIRI);
+    });
+</script>
```
stuzart, line 12: in the interest of trying to remove inline javascript, this block could be moved to sample_types.js . I don't see `term-label` used anywhere else, but to be more specific I'd use `label.term-label` as the selector.

stuzart (changes requested):

app/helpers/samples_helper.rb
whomingbird: tested, no problem without `.html_safe`

app/assets/javascripts/sample_types.js
whomingbird: image
if using `.show()` , it will show the term-iri in the next row.

app/helpers/samples_helper.rb
whomingbird: changed to `iri_content = term.iri.match?(/^https?:\/\//) ? link_to(term.iri, term.iri, target: '_blank') : term.iri`

app/assets/javascripts/sample_types.js
stuzart: OK, can leave it then and is not that important. I also realise now that it is toggling it rather than just showing it.

stuzart (approved):

PR #2001, opened 2024-10-02
Fix CWL parsing issue

fbacall:
...when the CWL has a `record` input/output with multiple fields.

Also fix an issue that arises when trying to remove the "main workflow path" of a workflow.


stuzart (approved):

stuzart (approved):
1 set of unit tests fell over, but the others were successful

PR #2003, opened 2024-10-03
Issue 2002 fix samples helper

### LINKED ISSUE #2002: undefined method `label' for nil:NilClass in method `seek_cv_attribute_display`
When sample routes are accessed, this error is thrown:

```
An ActionView::Template::Error occurred in samples#show:

  undefined method `label' for nil:NilClass

      term.label
          ^^^^^^
  app/helpers/samples_helper.rb:185:in `seek_cv_attribute_display'
```
Probably caused by:
https://github.com/seek4science/seek/blob/41337acfefd729435f4329eda12fe428459b433a/app/helpers/samples_helper.rb#L177-L187

kdp-cloud:
Small fix for CV attributes that allow free text.
Closes #2002

### DIFF
--- a/test/unit/helpers/samples_helper_test.rb
+++ b/test/unit/helpers/samples_helper_test.rb
@@ -86,16 +86,27 @@ class SamplesHelperTest < ActionView::TestCase
   test 'cv attribute display with free text' do
     person = FactoryBot.create(:person)
     User.with_current_user(person.user) do
-      attribute = FactoryBot.create(:apples_controlled_vocab_attribute,
-                                    sample_type: FactoryBot.create(:simple_sample_type),
-                                    allow_cv_free_text:true)
-      existing_term = attribute.sample_controlled_vocab.sample_controlled_vocab_terms.first.label
+      cv_attribute = FactoryBot.create(:apples_controlled_vocab_attribute,
+                                       sample_type: FactoryBot.create(:simple_sample_type),
+                                       allow_cv_free_text: true)
+      existing_cv_term = cv_attribute.sample_controlled_vocab.sample_controlled_vocab_terms.first.label
       free_text = 'new term'
 
-      [existing_term, free_text].each do |value|
+      [existing_cv_term, free_text].each do |value|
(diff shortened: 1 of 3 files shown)


test/unit/helpers/samples_helper_test.rb
```diff
@@ -82,4 +82,39 @@ class SamplesHelperTest < ActionView::TestCase
...
+    User.with_current_user(person.user) do
+      attribute = FactoryBot.create(:apples_controlled_vocab_attribute,
+                                    sample_type: FactoryBot.create(:simple_sample_type),
+                                    allow_cv_free_text:true)
+      existing_term = attribute.sample_controlled_vocab.sample_controlled_vocab_terms.first.label
+      free_text = 'new term'
+
+      [existing_term, free_text].each do |value|
+        assert_nothing_raised do
+          display_attribute_value(value, attribute)
```
stuzart, line 97: would be good if this also checked the return value. Even though an exception is no longer raised, it could be returning nil or something else incorrect instead.

test/unit/helpers/samples_helper_test.rb
```diff
@@ -82,4 +82,39 @@ class SamplesHelperTest < ActionView::TestCase
...
+      attribute = FactoryBot.create(:apples_list_controlled_vocab_attribute,
+                                    sample_type: FactoryBot.create(:simple_sample_type),
+                                    allow_cv_free_text:true)
+      existing_list = [attribute.sample_controlled_vocab.sample_controlled_vocab_terms.first.label,
+                       attribute.sample_controlled_vocab.sample_controlled_vocab_terms.second.label]
+      mixed_list = [existing_list[0], 'new term']
+
+      [existing_list, mixed_list].each do |value|
+        assert_nothing_raised do
+          display_attribute_value(value, attribute)
```
stuzart, line 115: likewise, it should check the return value

stuzart (changes requested):

whomingbird (approved):

stuzart (approved):

Issue #1999, opened 2024-09-30
 System Admin has right to create, manage (enable, disable), and delete extended metadata types via user interface 


whomingbird:
related [issue](https://github.com/seek4science/seek/issues/1747): Feature request: Management of Extended Metadata via Frontend #1747

- [x] as a SEEK instance administrator, you have the ability to create, manage (enable, disable), and delete extended metadata types. 

- [x] on the "Manage Extended Metadata Types" page, you will see a list of existing extended metadata types along with relevant resources for creating new ones.

- [x] you can upload a JSON file through the Admin Area UI to create the metadata type.

carlosvega:
I wonder about editing/updating extended metadata types. For instance, what's the recommended procedure if we want to add one new field to a extended metadata template?

PR #2010, opened 2024-10-09
System Admin has right to create, manage (enable, disable), and delete extended metadata types via user interface #1999

whomingbird:
https://github.com/seek4science/seek/issues/1999


app/models/extended_metadata_type.rb
```diff
@@ -34,6 +34,14 @@ def extended_type?
     supported_type == 'ExtendedMetadata'
   end
 
+  def linked_by_metadata_attributes?
+    ExtendedMetadataAttribute.pluck(:linked_extended_metadata_type_id).compact.include?(id)
```
stuzart, line 38: can this not just be `linked_metadata_attributes.any?` and avoid looping over all the attribute ids

app/views/extended_metadata_types/_emt_tabs.erb
```diff
@@ -0,0 +1,156 @@
...
+    </a>
+  </li>
+</ul>
+
+<div class="tab-content">
+  <div class="tab-pane fade active in" id="top-level-metadata-table" role="tabpanel" aria-labelledby="top-level-tab">
+    <div class="alert alert-info">
+      <ul>
+        <li>
+          You can add extended metadata to the following SEEK resource types:  <strong>Investigation</strong>, <strong>Study</strong>, <strong>Assay</strong>, <strong>Collection</strong>, <strong>DataFile</strong>, <strong>Document</strong>, <strong>Event</strong>, <strong>Model</strong>, <strong>Presentation</strong>, <strong>Sop</strong>, and <strong>Project</strong>.
```
stuzart, line 42: this should use `Seek::Config.instance_name` and the locale's for the type names. It could iterate over all the types with `supports_extended_metadata?` rather than hard coded

app/views/extended_metadata_types/_form.html.erb
```diff
@@ -0,0 +1,26 @@
+<%= form_tag extended_metadata_types_path, method: "post", id: "extended_metadata_type_upload_from", multipart: true do -%>
+  <%= folding_panel("JSON file", false, :body_options => {:id => 'add_extended_metadata_type_file_fold_content'},
```
stuzart, line 2: avoid using the old `:x=>` but use `x:` instead

app/views/extended_metadata_types/_help_info.erb
```diff
@@ -0,0 +1,8 @@
+<!-- the links will be updated after the official doc is available -->
+<div>
+  For more information, please visit our
+  <%= link_to 'Extended Metadata Technical Overview', 'https://github.com/seek4science/seek-documentation/blob/gh-pages-extended_metadata-type/tech/extended_metadata/extended-metadata-type.md', target: '_blank' %> page.
+  This is <%= link_to 'JSON schema', 'https://github.com/seek4science/seek-documentation/blob/gh-pages-extended_metadata-type/tech/extended_metadata/extended_metadata_type_schema.json', target: '_blank' %> used to define your JSON file for creating a new extended metadata type via upload.
+  Here’s a <%= link_to 'step-by-step example', 'https://github.com/seek4science/seek-documentation/blob/gh-pages-extended_metadata-type/tech/extended_metadata/a-complete-example.md', target: '_blank' %> of how to create a new extended metadata type.
```
stuzart, line 6: use the help_dictionary for these links, where they can then also be updated when they go live

lib/seek/extended_metadata_type/extended_metadata_type_schema.json
```diff
@@ -0,0 +1,120 @@
...
+        "Assay",
+        "Collection",
+        "DataFile",
+        "Document",
+        "Event",
+        "Model",
+        "Presentation",
+        "Sop",
+        "Project",
+        "ExtendedMetadata"
```
stuzart, line 22: I don't these should go in the schema, but instead check programmatically that the type exists and `supports_extended_metadata?`

lib/seek/extended_metadata_type/extended_metadata_type_schema.json
```diff
@@ -0,0 +1,120 @@
...
+                  "MetaNetX reaction",
+                  "MetaNetX compartment",
+                  "InChI",
+                  "ENA custom date",
+                  "Boolean",
+                  "URI",
+                  "DOI",
+                  "NCBI ID",
+                  "Registered Strain",
+                  "Registered Data file"
```
stuzart, line 65: these can't be validated through the schema, as it depends on how the server is configured and people may have added their own or removed some.

stuzart (changes requested):

lib/seek/extended_metadata_type/extended_metadata_type_schema.json
whomingbird: I believe the schema should include the supported types. This not only serves as validation for the allowed types but also helps users when creating their JSON files. 
Would checking the supported types programmatically create duplication if they're already defined in the schema?

whomingbird: I have added a validation in [ExtendedMetadataType model ](https://github.com/seek4science/seek/blob/0c653980d2484513d87a701946900340c8b0efd5/app/models/extended_metadata_type.rb#L10) , but i think we should not remove them from schema.

app/models/extended_metadata_type.rb
```diff
@@ -62,6 +52,16 @@ def cannot_disable_nested_extended_metadata
     end
   end
 
+  def supports_extended_metadata
+    begin
+      unless self.supported_type.constantize.supports_extended_metadata?
```
stuzart, line 57: use `Seek::Util.lookup_class` instead of `constantize` as it's safer and more secure

lib/seek/util.rb
```diff
@@ -180,5 +180,11 @@ def self.filter_disabled(types)
       types.select(&:feature_enabled?)
     end
 
+    def self.extended_metadata_supported_types
+      ActiveRecord::Base.descendants.select do |model|
```
stuzart, line 184: you need to be sure the models have been loaded by the classloader first. You can do this by using `persistent_classes`. Also cache. Have a look at the other methods, like authorized_types - the pattern is the same just calling a different method

lib/seek/util.rb
```diff
@@ -180,5 +180,11 @@ def self.filter_disabled(types)
...
+        model.respond_to?(:supports_extended_metadata?) && model.supports_extended_metadata?
+      end.map(&:name)
```
stuzart, line 186: return the types, instead of their String names - to be consistent with the other Util methods

stuzart (changes requested):
just a couple of minor things

lib/seek/util.rb
stuzart: by which I mean the classes

stuzart (approved):

Issue #1981, opened 2024-08-28
Add the ability to update an investigation with a new TTL file from FAIR data station.


stuzart:
From an investigation page, which has been created from FAIR data station, add the ability to upload a new TTL file to update that investigation and its component Studies, Obs Units, Samples and Assays.
Without changing the extended metadata type, it will update any attributes for each item, currently matched by external id. It will create new entries where an existing item cannot be found. I won't delete missing entries.

- [ ] Add a flag to mark when an investigation has been created via fair data station. Also add this flag to study, obs unit, sample and assay.
- [ ] Create a test case that includes: changed metadata for each type, a new study, obs unit, sample, assay chain, and new sample to existing obs unit, a new data set reference.
- [ ] Menu option on the Investigation to Update from FAIR data station (if enabled)
- [ ] Find matching entry by type and title, and update existing entry attributes where the PID matches, clear previous attributes where they don't
- [ ] Create new entries if not matched

PR #2017, opened 2024-10-14
Merging from FAIR Data Station

stuzart:
Supports updating Investigation->Study->ObsUnit->Sample->Assays with new information from FAIR Data Station.

Supports

- Changing existing metadata
- Moving items, e.g an Observation Unit to a different Study
- Creating new items

It doesn't currently support deleting items, as there may be cases where things are being updated from a new template with partial information (like a PATCH). An option to delete may be added in the future.

* fix for #1981


app/views/investigations/update_from_fairdata_station.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render :partial => "general/page_title",:locals=>{:title=>"Update from FAIR Data Station for #{ t('investigation') } #{link_to(@investigation.title, @investigation)}".html_safe} %>
```
whomingbird, line 1: use new style, not :x =>

app/views/projects/import_from_fairdata_station.html.erb
```diff
@@ -1,4 +1,24 @@
-<%= render :partial => "general/page_title",:locals=>{:title=>"Import ISA from FAIRData Station for #{ t('Project') } #{link_to(@project.title, @project)}".html_safe} %>
+<%= render :partial => "general/page_title",:locals=>{:title=>"Import ISA from FAIR Data Station for #{ t('Project') } #{link_to(@project.title, @project)}".html_safe} %>
```
whomingbird, line 1: still the style

lib/seek/fair_data_station/writer.rb
```diff
@@ -33,7 +53,14 @@ def reset_data_file_cache
         @data_file_cache = {}
       end
 
-      def build_assay(datastation_assay, contributor, policy, projects, sample, study)
+      def preload_data_file_cache(data_files)
```
whomingbird, line 56: question: why do you need to do this?

lib/seek/fair_data_station/writer.rb
```diff
@@ -1,24 +1,44 @@
...
+                build_assay(datastation_assay, contributor, projects, policy, sample, study)
               end
+            end
+          end
+        end
+
+        investigation
+      end
+
+      def update_isa(investigation, datastation_inv, contributor, projects, policy)
```
whomingbird, line 26: I understand that the current `update_isa` method only supports fair_data_station resource types derived from SEEK resources. In the future, could this method be adapted for use with SEEK ISA elements as well?  A good use case could be creating and updating a whole ISA through the API.

whomingbird (approved):
The changes look solid, and I’ve tested them through the user interface as well. One suggestion for future work: it would be beneficial to enhance the response after 'update' or 'move' actions, so users can have a clear overview of what has been changed.

lib/seek/fair_data_station/writer.rb
stuzart: data files can be connected in multiple places, so may be encountered more than once. The cache is used to preven't multiple duplicates of the data files being created. For the import this is built up whilst as they are created, but for the update the cache needs preloading, otherwise new ones will start being created again.

lib/seek/fair_data_station/writer.rb
stuzart: yes, although thinking more in terms of supporting RO-Crates. The FAIR data station TTL is very similar to what an RO-Crate may look like.
For SEEK to SEEK migration the UUID's may be safer to use than the external identifiers, it's what they were originally added for.

stuzart:
> The changes look solid, and I’ve tested them through the user interface as well. One suggestion for future work: it would be beneficial to enhance the response after 'update' or 'move' actions, so users can have a clear overview of what has been changed.

thanks. Yes, the UI of the results is something I also think will be needed, but as a later task. I'm also not sure how much this will be used by a human rather than via an API.
I think also it will eventually needed to be turned into background job, but waiting to get a better idea of the likely size of the imports and updates

PR #2031, opened 2024-10-21
Remove duplicate table

kdp-cloud:
Removes duplicated `sample_type_auth_lookup` table.


stuzart (approved):
not quite a duplicate, the one being removed it pluralized `sample_type_auth_lookups`.
Checked, and both were added with the same commit and must have slipped through the pull request.

PR #2032, opened 2024-10-21
362 add and edit attributes in sample types

### LINKED ISSUE #362: Add and edit attributes
**User Story ID: 12.0**
As PhD student, I want to add and edit attributes to my Study and Assays any time I need to, so that I can edit attributes that were wrong or add attributes that I forgot because they were not important before.

**Acceptance Criteria**

**Definition of Done**

### LINKED ISSUE #1257: Allow for addition of optional Attributes to existing Sample Types
As discussed in the Samples WG, adding optional Attributes to existing Sample Types should be allowed.

Following issue #1199 , the pool of users who can edit a Sample Type will be more clear.

### LINKED ISSUE #1258: Allow editing of Attributes in existing Sample Types
As discussed in the Samples WG.

For existing Sample Types. Allow for edits that do not lead to validation conflicts with existing Samples.
Allow the following changes:
- An Attribute from required to optional
- The Attribute tagged as Title
- Attribute Name
- Attribute PID
- Attribute Description
- Attribute Unit

kdp-cloud:
- Adds possibility to add new optional attribute to an existing sample type
- Makes the title of a sample attribute editable.
- If the title of an attribute is updated, the JSON metadata values of all samples that are linked to the sample type are also updated as a background job.

Closes #1257 
Closes #1258 
Closes #362

### DIFF
--- a/app/controllers/sample_types_controller.rb
+++ b/app/controllers/sample_types_controller.rb
@@ -13,6 +13,7 @@ class SampleTypesController < ApplicationController
   before_action :auth_to_create, only: %i[new create]
   before_action :project_membership_required, only: %i[create new select filter_for_select]
 
+  after_action :update_sample_json_metadata, only: :update
 
   api_actions :index, :show, :create, :update, :destroy
 
@@ -200,5 +201,9 @@ class SampleTypesController < ApplicationController
       flash[:error] = "Cannot #{action_name} this sample type - There are #{count} samples using it."
       redirect_to @sample_type
     end
+  end
+
+  def update_sample_json_metadata
+    UpdateSampleMetadataJob.perform_later(@sample_type)
   end
 end
--- a/app/controllers/sample_types_controller.rb
+++ b/app/controllers/sample_types_controller.rb
@@ -227,6 +227,7 @@ class SampleTypesController < ApplicationController
 
   def check_if_locked
     @sample_type ||= SampleType.find(params[:id])
+    @sample_type.reload
     return unless @sample_type&.locked?
 
     error_message = 'This sample type is locked and cannot be edited right now.'
--- a/app/controllers/sample_types_controller.rb
+++ b/app/controllers/sample_types_controller.rb
@@ -204,6 +204,6 @@ class SampleTypesController < ApplicationController
   end
 
   def update_sample_json_metadata
-    UpdateSampleMetadataJob.perform_later(@sample_type)
+    UpdateSampleMetadataJob.perform_later(@sample_type) unless @sample_type.samples.blank?
   end
 end
--- a/app/controllers/sample_types_controller.rb
+++ b/app/controllers/sample_types_controller.rb
@@ -207,16 +207,20 @@ class SampleTypesController < ApplicationController
   def old_attributes
     return if @sample_type.sample_attributes.blank?
 
-    @old_sample_type_attributes = @sample_type.sample_attributes
+    @old_sample_type_attributes = @sample_type.sample_attributes.map { |attr| { id: attr.id, title: attr.title } }
   end
 
   def update_sample_json_metadata
     return if @sample_type.samples.blank? || @old_sample_type_attributes.blank?
 
-    attribute_changes = @sample_type.sample_attributes.each do |attr|
-      old_attr = @old_sample_type_attributes.detect { |oa| oa.id == attr.id }
-      { id: attr.id, old_title: old_attr&.title, new_title: attr.title } unless old_attr&.title == attr.title
-    end
-    UpdateSampleMetadataJob.perform_later(@sample_type, attribute_changes) unless attribute_changes.blank?
+    attribute_changes = @sample_type.sample_attributes.map do |attr|
+      old_attr = @old_sample_type_attributes.detect { |oa| oa[:id] == attr.id }
+      next if old_attr.nil?
+
+      { id: attr.id, old_title: old_attr[:title], new_title: attr.title } unless old_attr[:title] == attr.title
+    end.compact
+    return if attribute_changes.blank?
+
+    UpdateSampleMetadataJob.perform_later(@sample_type, attribute_changes, @current_user)
   end
 end
--- a/app/controllers/sample_types_controller.rb
+++ b/app/controllers/sample_types_controller.rb
@@ -6,10 +6,10 @@ class SampleTypesController < ApplicationController
 
   before_action :samples_enabled?
   before_action :check_no_created_samples, only: [:destroy]
+  before_action :check_if_locked, only: %i[edit manage manage_update update]
   before_action :find_and_authorize_requested_item, except: %i[create batch_upload index new template_details]
   before_action :find_sample_type, only: %i[batch_upload template_details]
   before_action :check_isa_json_compliance, only: %i[edit update manage manage_update]
-  before_action :check_if_locked, only: %i[edit manage update manage_update]
   before_action :find_assets, only: [:index]
   before_action :auth_to_create, only: %i[new create]
   before_action :project_membership_required, only: %i[create new select filter_for_select]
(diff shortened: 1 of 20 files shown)


lib/seek/samples/sample_type_editing_constraints.rb
```diff
@@ -49,24 +49,17 @@ def allow_attribute_removal?(attr)
...
-          return false if inherited?(attr)
 
           attr = attr.accessor_name
         end
-        if attr
-          !samples?
-        else
-          true
-        end
+        return !samples? unless attr
```
fbacall, line 59: Is this logic correct? Previously it would allow name changes if attr was nil

app/jobs/update_sample_metadata_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class UpdateSampleMetadataJob < ApplicationJob
+  queue_with_priority 1
+  queue_as QueueNames::SAMPLES
+
+  def perform(sample_type, attribute_changes = [], user)
+    Seek::Samples::SampleMetadataUpdater.new(sample_type, attribute_changes, user).update_metadata
```
fbacall, line 8: General question, but what happens if someone views a sample before the job has finished running?

app/views/isa_studies/_sample_types_form.html.erb
```diff
@@ -52,7 +52,7 @@
                                                                                   display_isa_tag: true } %>
 					<% end %>
 
-					<% unless sample_type.uploaded_template? || !sample_type.editing_constraints.allow_new_attribute? %>
```
fbacall, line 55: I think I would leave these checks in the code, but make the `allow_new_attribute?` method return `true`, just in case we want to change the behaviour in the future.

lib/seek/samples/sample_type_editing_constraints.rb
```diff
@@ -49,24 +49,17 @@ def allow_attribute_removal?(attr)
         end
       end
 
-      # whether a new attribtue can be created
-      def allow_new_attribute?
-        !samples?
-      end
```
fbacall, line 55: See previous comment

lib/seek/samples/sample_metadata_updater.rb
```diff
@@ -0,0 +1,32 @@
...
+        @sample_type = sample_type
+        @attribute_change_maps = attribute_changes
+        @user = user
+      end
+
+      def update_metadata
+        return if @attribute_change_maps.blank? || @sample_type.samples.blank? || @sample_type.nil? || @user.nil?
+
+        User.with_current_user(@user) do
+          @sample_type.samples.each do |sample|
```
fbacall, line 20: I think this may have to work in batches, as updating thousands of samples could take a long time and the job may time out (as has happened with deletes)

app/controllers/sample_types_controller.rb
```diff
@@ -12,7 +12,9 @@ class SampleTypesController < ApplicationController
   before_action :find_assets, only: [:index]
   before_action :auth_to_create, only: %i[new create]
   before_action :project_membership_required, only: %i[create new select filter_for_select]
+  before_action :old_attributes, only: %i[update]
 
+  after_action :update_sample_json_metadata, only: :update
```
fbacall, line 17: What would happen if 2 updates were made to a sample type in a short time?

I think you may need a mechanism for "locking" the sample type to further changes until the update job has completed.

lib/seek/samples/sample_metadata_updater.rb
```diff
@@ -0,0 +1,32 @@
...
+            metadata = JSON.parse(sample.json_metadata)
+            # Update the metadata keys with the new attribute titles
+            @attribute_change_maps.each do |change|
+              metadata[change[:new_title]] = metadata.delete(change[:old_title])
```
fbacall, line 24: If someone creates a new sample before the job finishes, this line will wipe out the changed attributes.

fbacall (commented):
Flagged a few issues, mostly around concurrency

lib/seek/samples/sample_type_editing_constraints.rb
kdp-cloud: If it has to return `true` anyway and we don't care about sample permissions, I guess this method can go because it would always return `true`.

app/jobs/update_sample_metadata_job.rb
kdp-cloud: Is there a way to test this?

app/controllers/sample_types_controller.rb
kdp-cloud: Didn't know that was possible. Has it already been done somewhere else in the source code?

app/jobs/update_sample_metadata_job.rb
fbacall: On your local instance you could update a sample type without any workers running. Alternatively, add a `sleep` here or in the job.

kdp-cloud:
I added a locking mechanism, locking the sample type and its linked samples while the background job is running. Creating new samples in this sample type is now prohibited as well.

app/models/sample_type.rb
```diff
@@ -156,7 +162,18 @@ def self.can_create?
...
+    super && !locked?
+  end
+
+  def can_manage?(user = User.current_user)
+    super && !locked?
+  end
+
   def can_delete?(user = User.current_user)
+    # Should not be able to delete a sample type if it is locked
+    return false if locked?
```
fbacall, line 175: I think these should be done using the `state_allows_edit?` etc. checks

e.g. https://github.com/seek4science/seek/blob/0995d8634d2da1ae0ac77851ecd36a467f94d25d/app/models/assay.rb#L136-L138

app/models/sample.rb
```diff
@@ -40,6 +40,7 @@ class Sample < ApplicationRecord
 
   validates_with SampleAttributeValidator
   validate :validate_added_linked_sample_permissions
+  validate :check_if_locked_sample_type, on: :create
```
fbacall, line 43: Does it also need check on update?

app/controllers/samples_controller.rb
```diff
@@ -8,6 +8,7 @@ class SamplesController < ApplicationController
   before_action :samples_enabled?
   before_action :find_index_assets, only: :index
   before_action :find_and_authorize_requested_item, except: [:index, :new, :create, :preview]
+  before_action :check_if_locked_sample_type, only: [:edit, :new]
```
fbacall, line 11: Also on `:update` and `:create`?

app/jobs/update_sample_metadata_job.rb
```diff
@@ -0,0 +1,18 @@
...
+class UpdateSampleMetadataJob < TaskJob
+  queue_with_priority 1
+  queue_as QueueNames::SAMPLES
+
+  def perform(sample_type, user, attribute_changes = [])
+    @sample_type = sample_type
+    @user = user
+    @attribute_changes = attribute_changes
+
+    Seek::Samples::SampleMetadataUpdater.new(@sample_type, @user, @attribute_changes).update_metadata
```
fbacall, line 12: I think perhaps it needs to queue a `SampleTypeUpdateJob` after updating the metadata, to refresh the cache on which attributes are blank etc.

test/functional/sample_types_controller_test.rb
```diff
@@ -5,6 +5,7 @@ class SampleTypesControllerTest < ActionController::TestCase
   include AuthenticatedTestHelper
 
   setup do
+    Rails.cache.clear
```
fbacall, line 8: Is this needed? The cache should be cleared between tests automatically

app/jobs/update_sample_metadata_job.rb
kdp-cloud: I thought it already did
https://github.com/seek4science/seek/blob/e824fcfcca42450fb2ad31f78164150f4d65b546/lib/seek/templates/sample_type_template_concerns.rb#L8

https://github.com/seek4science/seek/blob/e824fcfcca42450fb2ad31f78164150f4d65b546/lib/seek/templates/sample_type_template_concerns.rb#L28-L30

fbacall (approved):

Issue #2030, opened 2024-10-18
ActivityLogs needed for FAIR Data Station import and update


stuzart:
currently not being written for update and create, as not going through the usual controller actions

PR #2046, opened 2024-11-06
FAIR Data Station import & update activity logs creation

stuzart:
Updates and tests to support generating ActivityLog records with 'create' action during importing from FAIR Data Station, and logs with 'update' action for items that change (or create for new items) during updating from FAIR Data Station.

Logs are only created with `update` only if their metadata changes (including extended metadata), but not if they only move. 

* fix for #2030


whomingbird (approved):
The code looks good to me.

Issue #1965, opened 2024-07-23
Search Function Issue


anneelinh:
**Describe the bug**
The search function on the page does not return results when typing and trying to search for "Thy-1". This should give a result as it is a part of the header of this publication (https://seek.lisym.org/publications/431). However, if you copy paste the the word from the publication it can find it. So there seems to be a problem with different hyphens being used.

**To Reproduce**
1. Enter "Thy-1" in the search bar and press enter.
2. Note that no results are returned.
3. Copy/paste "Thy‐1" from the header of the publication (see above) in the search bar and press enter.
4. Note that relevant results are returned.

**Expected behavior**
The search function should return the same results for both typing "Thy-1" and for copy/pasting "Thy‐1" from the publication.

**Screenshots**
![image](https://github.com/user-attachments/assets/f69bb0cc-dd5d-4ccf-9db0-80ee9ceb8b0f)
![image](https://github.com/user-attachments/assets/38eb43ed-58ee-4f56-9c3a-87ba17fc4850)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 126.0.6478.182

stuzart:
The hyphen `-` in the title is not the same character as the '-' you would type, it's a funny unicode character. It's actually `Thy&#x2010;1`

PR #2072, opened 2024-11-22
Supporting Special Characters in Unicode for Search Functions

whomingbird:
https://github.com/seek4science/seek/issues/1965


whomingbird:
> is this intended for 1.16.0 ?
yes.

solr/seek/conf/mapping-hyphen.txt
```diff
@@ -0,0 +1,7 @@
+# Mapping of Unicode hyphen to ASCII hyphen
+"\u2010" => "-"
+"\u2011" => "-"
+"\u2012" => "-"
+"\u2013" => "-"
+"\u2014" => "-"
+"\u2015" => "-"
```
stuzart, line 7: I don't this file is needed, as they are already included in teh FoldToASCII

test/functional/search_controller_test.rb
```diff
@@ -155,6 +155,58 @@ class SearchControllerTest < ActionController::TestCase
     assert_select 'a[href="#test-123"]', count: 0
   end
 
+
+  test 'hyphen and unicode hyphen are treated equally in search' do
```
stuzart, line 159: you realise this isn't testing the search, as it is stubbed out ?
but still useful I think for checking the flash notice text contains the original text.

stuzart (changes requested):

stuzart:
The docker container at https://github.com/FAIRdom/solr-seek-docker also needs updating. If you update this too, on a branch, we can give a test on the test server before merging

solr/seek/conf/mapping-hyphen.txt
whomingbird: i just realized that

stuzart (approved):

Issue #2070, opened 2024-11-20
specialize MailDeliveryJob to pick up changed smtp settings without needing to restart job


stuzart:
The job used is configured here: https://github.com/seek4science/seek/blob/main/config/application.rb#L75

would just need to subclass `MailDeliveryJob`, and then propogate the smtp settings before the job is performed - similar to [EmailJob](https://github.com/seek4science/seek/blob/main/app/jobs/email_job.rb)

PR #2075, opened 2024-11-25
MailDeliveryJob to updates with settings and avoids needing to restart.

stuzart:
Extension of MailDeliveryJob to update smtp settings before running.
Also always start the mailer job, but only send mails if enabled, to avoid having to restart the jobs if enabling email. ( I'm in 2 minds about this and open to reverting it after review. )
Added a note that when queuing a test email, the settings need to have been saved first.

* Fix for #2070

### DIFF
--- a/app/views/admin/_email.html.erb
+++ b/app/views/admin/_email.html.erb
@@ -55,6 +55,7 @@
         <%= link_to '#', id: 'queue-test-email-btn', class: 'btn btn-default' do %>
           <%= image("test", title: "Send testing email", alt: "Test")%> Queue test email
         <% end %>
+        <p class="help-block">Note that testing the Queue requires settings to be saved first</p>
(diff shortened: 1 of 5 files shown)


app/views/admin/_email.html.erb
```diff
@@ -55,6 +55,7 @@
         <%= link_to '#', id: 'queue-test-email-btn', class: 'btn btn-default' do %>
           <%= image("test", title: "Send testing email", alt: "Test")%> Queue test email
         <% end %>
+        <p class="help-block">Note that testing the Queue requires settings to be saved first</p>
```
fbacall, line 58: Perhaps in the future we could make the custom job take an optional hash of SMTP settings to temporarily override the saved settings, so it can be used for testing without needing to save the settings first.

PR #2081, opened 2024-11-29
use snake_case to replace spaces in the title to create the ID for Ja…

whomingbird:


fbacall (commented):
There is a Rails helper method `dom_id` for generating safe IDs that can be used in HTML/JS

https://api.rubyonrails.org/v7.1/classes/ActionView/RecordIdentifier.html#method-i-dom_id

Maybe you could use `dom_id(attribute)` or something?

fbacall (approved):

Issue #2084, opened 2024-11-29
Download sampletype template is shown and available if can_download? is false


stuzart:
Previously the Template could always be downloaded if the sample type was visible, but since the introduction of permissions can_download? has been added as an option. As discovered with #2076 the download button is shown, but only if there is permissions, but the download link on the sample type view page is always shown and accessible if visible.

should also make sure and add tests that the action is blocked if can_download? is false.

There is a question of whether this separate permission is needed, as downloading the template doesn't reveal anything not already known.

PR #2085, opened 2024-12-02
Sample Type and download permissions 

### LINKED ISSUE #2076: NoMethodError for downloading Sample Types
Accessing `/sample_types/:id/download` raises a NoMethodError:

```
A NoMethodError occurred in sample_types#download:

  undefined method `content_blob' for nil:NilClass

                          [asset_version.content_blob]
                                        ^^^^^^^^^^^^^
  lib/seek/content_blob_common.rb:15:in `download'
```

stuzart:
Override methods to prevent sample types being treated as a downloadable asset. This means it behaves more like Samples, Investigation, Studies and Assays, and the permission options don't include a Download option (however, the ACCESSIBLE access type is applied).

Also updated the upgrade task to apply ACCESSIBLE instead of VIEWABLE, and a new task to update the previously applied permissions.

Although unlikely to be the case, the Download template option isn't shown if viewable but not accessible.

* addresses #2084 
* a side effect also fixed #2076 as the asset actions, with the download button, are no longer shown for list items.

### DIFF
--- a/app/models/sample_type.rb
+++ b/app/models/sample_type.rb
@@ -200,6 +200,19 @@ class SampleType < ApplicationRecord
     can_view?(user)
   end
 
+  def self.is_asset?
+    false
+  end
+
+  # although has a downloadable template, it doesn't have the full downloadable behaviour of an asset with data and it's own accessible permissions
+  def is_downloadable?
+    false
+  end
+
+  def self.supports_extended_metadata?
+    false
+  end
+
   private
 
   # whether the referring sample is valid and gives permission to view
@@ -290,10 +303,6 @@ class SampleType < ApplicationRecord
     if is_title_seek_sample_multi
       errors.add(:sample_attributes, "Attribute type of #{base_type.underscore.humanize} can not be selected as the sample type title.")
     end
-  end
-
-  def self.supports_extended_metadata?
-    false
   end
 
   class UnknownAttributeException < RuntimeError; end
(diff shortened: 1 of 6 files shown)


app/models/sample_type.rb
```diff
@@ -292,9 +305,5 @@ def validate_title_is_not_type_of_seek_sample_multi
     end
   end
 
-  def self.supports_extended_metadata?
-    false
-  end
-
```
kdp-cloud, line 298: Just a question. Does it support Extended Metadata now? Or is it not necessary  anymore because it's an asset now?

lib/tasks/seek_upgrades.rake
```diff
@@ -125,7 +126,22 @@ namespace :seek do
...
+    puts "... Added policies to #{counter} sample types"
+  end
+
+  task(fix_previous_sample_type_permissions: [:environment]) do
+    only_once('fix_previous_sample_type_permissions 1.16.0') do
+      puts '... Updating previous sample type permissions ...'
+      SampleType.includes(:policy).where.not(policy_id: nil).each do |sample_type|
+        policy = sample_type.policy
+        if policy.access_type == Policy::VISIBLE
+          policy.update_column(:access_type, Policy::ACCESSIBLE)
```
kdp-cloud, line 138: Again a question. Does a sample type need to be accessible in order to be able to download the template?

kdp-cloud (approved):

app/models/sample_type.rb
stuzart: SampleTypes don't support EM. I didn't add this by the way, but just moved up the class out of the private area so I could group these overrides together

lib/tasks/seek_upgrades.rake
stuzart: Yes, because it goes throught the content blob controller it checks the permissions on the associated asset. There is an example here: https://testing.sysmo-db.org/sample_types/97

PR #2104, opened 2025-01-08
Issue 1147 fix cv duplication in template json import

### LINKED ISSUE #1147: CV title
**Describe the bug**
When creating a new sample_controlled_vocabs (in both Template and Sample Type), if I give an existing CV title SEEK gives warning error, since CV titles must be unique and no duplicates are allowed.
However, if I look at the list of existing CV titles I see many duplicates (e.g., library source is listed 4 times at least).

**To Reproduce**
See screenshot below and then:

1. Go to Create --> Template or Sample type
2. Click on Attribute Type --> Controlled Vocabulary or Ontology --> use dropdown to see the existing ones
3. Scroll down and see titles such as "library source" listed multiple times.

This is in contrast with the following error:
Screenshot 2022-08-04 at 14 11 11

**Expected behavior**
Consistent behaviour, although I don't know yet what that would be. To be discussed.

**Desktop (please complete the following information):**

**Additional context**
In DataHub, the same attribute (like, library source) with the same CV list (titled library source as well) will be "used" many many times by derived templates (via Create Template based on existing templates) and derived sample types (via applying the same template to many studies or assays). So, I think this is what it is causing having repeated CV list titles in the dropdown menu.
However, this is in contrast with default SEEK rule that prohibits having repetition of the same title in CV list.

Questions:
- does SEEK only check if the CV title exist, independently of the listed terms being identical or different? probably, yes.
- should the content of the list be checked?
- is this a big limit for users since all CV lists are listed together, meaning CV lists created by anyone in any project at any point in time in a given SEEK instance?
- should lists be also restricted somehow (by project, by template...)?
- should lists make use of unique identifiers (seek ID) to check unicity?
- how does DataHub differentiate between Ontology and CV? Currently, the GUI to create an Ontology list is identical to the one for creating a CV list.
- It might be necessary now to define what's an Ontology list vs a CV list
  - Ontology list:
    - only one root ontology term and all children are fetched
    - no editing allow (no deletion, no addition of free terms)
    - Ontology list take the name of the root term as title
  - CV list:
    - no direct fetching of ontology terms from OLS (but still possible to paste URL to a ontology term)
    - possibility to add, delete, edit terms
    - free text for CV list title

kdp-cloud:
- Reuses Controlled vocabularies instead of creating duplicates every time a new instance-wide template is uploaded.
- Add control of the `allow_cv_free_text` attribute in template attributes.
- Closes #1147


lib/seek/isa_templates/template_attributes_schema_test.json
stuzart, line 1: why is there a separate test schema? If it's somehow for the tests, would these not be better be testing against the offical schema ?

lib/seek/isa_templates/template_attributes_schema.json
```diff
@@ -1,112 +1,140 @@
...
+        },
+        {
+          "type": "null"
+        }
+      ]
+    },
+    "dataType": {
+      "type": "string",
+      "enum": [
+        "Date time",
```
stuzart, line 69: these attribute types aren't necessarily locked down, and can be changed or new ones added through the database - meaning the schema will need change, which would be readonly in Docker. Not a blocker for this PR, but something to concider changing. This is something that needed changing for the extended metadata type schema, and it would be worth trying to align them

stuzart (approved):
Approved, but with a couple of comments

lib/seek/isa_templates/template_attributes_schema_test.json
kdp-cloud: It's because the sample attribute type is sometimes created from a factory or sometimes reused from a fixture.
This means that  identical attributes will have a different titles, which I need to fixed in order to test the template upload of a JSON file.

Example:
If you create an an attribute from a factory, the attribute type will be something like 'String attribute type 5', depending on the number of tests creating a string type attribute. For me though, it would be easier if the name remains consistent (fixtures).

But I think this would be a good opportunity to simplify this indeed and only test one JSON schema. I created an issue: #2108

lib/seek/isa_templates/template_attributes_schema.json
kdp-cloud: I agree! It already happened a couple of times...

PR #2107, opened 2025-01-08
Implement secure cookies by default

kdp-cloud:
- Use for secure cookies in  a production environment
- Force SSL in production environment

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -99,7 +99,7 @@ class SessionsController < ApplicationController
     flash[:notice] = notice || "You have successfully logged in, #{@user.display_name}."
     if params[:remember_me] == 'on'
       @user.remember_me unless @user.remember_token?
-      cookies[:auth_token] = { value: @user.remember_token, expires: @user.remember_token_expires_at }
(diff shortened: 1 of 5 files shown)


stuzart (commented):
I'm not clear how this will affect running with docker and an nginx / apache front end forwarder. Will need some testing, possible ngnix configuration changes, and maybe an ENV variable to override. 
For these reasons I think it might be better left for the `main` branch and next major version

kdp-cloud:
> I'm not clear how this will affect running with docker and an nginx / apache front end forwarder. Will need some testing, possible ngnix configuration changes, and maybe an ENV variable to override. For these reasons I think it might be better left for the `main` branch and next major version

I changed the branch to main and the milestone to 1.17, but I'm not sure you should override this config in production since this implicates security issues.

kdp-cloud:
As discussed:
- HSTS directive will be set on proxy level
- Secure cookies will be set on proxy level
- Enabling HSTS and secure cookies will described in the documentation ([PR](https://github.com/seek4science/seek-documentation/pull/7))
- The SameSite directive is defined application-side:
  - The session cookie is set to 'Lax' to permit external authentication like OAUTH2
  - Other cookies are set to 'Strict'

PR #2109, opened 2025-01-10
Better message for querying samples

### LINKED ISSUE #2096: Better result message when querying samples
When querying samples and no result is returned, it is not clear that nothing was found. It should be clear that the user's query did not return any result

kdp-cloud:
- Doesn't fail if attribute is not selected
 - Provides a message when a query fails
 - Provides an error message when an error occurs
 - Spinner is added to indicate a running query
 - closes #2096

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -247,7 +247,7 @@ class SamplesController < ApplicationController
       end
     end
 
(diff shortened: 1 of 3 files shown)


kdp-cloud:
@NielsGeudens Some screenshots:

A running query
image

A successful query
image

An error occurs
image

stuzart (dismissed):

PR #2112, opened 2025-01-15
Upgrade to Rails 7.2 and Ruby 3.3

stuzart:
All the changes required for the upgrade. 
No need for a full details review of this large PR, but check for unexpectedly added files and silly errors.


app/models/isa_tag.rb
```diff
@@ -1,8 +1,8 @@
-class IsaTag < ApplicationRecord
+class ISATag < ApplicationRecord
   validates :title, presence: true
 
-  has_many :template_attributes, inverse_of: :isa_tag
-  has_many :sample_attributes, inverse_of: :isa_tag
+  has_many :template_attributes, inverse_of: :ISATag
+  has_many :sample_attributes, inverse_of: :ISATag
```
fbacall, line 5: I think these `inverse_of`s were accidentally changed, they should match the association name in the `TemplateAttribute` etc. models

app/models/project.rb
```diff
@@ -45,7 +45,7 @@ class Project < ApplicationRecord
 
   has_annotation_type :funding_code
 
-  enum status: [:planned, :running, :completed, :cancelled, :failed]
+  #enum status: [:planned, :running, :completed, :cancelled, :failed]
```
stuzart, line 48: this can be deleted, rather than commented out

app/models/assay.rb
```diff
@@ -1,7 +1,7 @@
 class Assay < ApplicationRecord
   include Seek::Ontologies::AssayOntologyTypes
 
-  enum status: %i[planned running completed cancelled failed]
+  #enum status: %i[planned running completed cancelled failed]
```
stuzart, line 4: delete

app/models/investigation.rb
```diff
@@ -10,7 +10,7 @@ class Investigation < ApplicationRecord
 
   validates :projects, presence: true, projects: { self: true }
 
-  enum status: [:planned, :running, :completed, :cancelled, :failed]
+  #enum status: [:planned, :running, :completed, :cancelled, :failed]
```
stuzart, line 13: delete

app/models/study.rb
```diff
@@ -1,6 +1,6 @@
 class Study < ApplicationRecord
 
-  enum status: [:planned, :running, :completed, :cancelled, :failed]
+  #enum status: [:planned, :running, :completed, :cancelled, :failed]
```
stuzart, line 3: delete

PR #2115, opened 2025-01-16
Issue 2106 support non text based attributes in dynamic table

### LINKED ISSUE #2106: Registered SOP / DataFile / Strain for ISA-JSON compliant sample types
ISA-JSON compliant sample types currently do not support these non-text-based types and need to be implemented.

kdp-cloud:
- Basic implementation of these attribute types in the dynamic table:
  - Registered SOP
  - Registered Datafiles
  - Registered Strains
- Closes #2106

### DIFF
--- a/app/controllers/sops_controller.rb
+++ b/app/controllers/sops_controller.rb
@@ -61,9 +61,9 @@ class SopsController < ApplicationController
 
     query = params[:query] || ''
     asset = if params[:study_id].present?
-              Study.find(params[:study_id]).authorized_for('view')
+              Study.authorized_for('view').detect { |study| study.id.to_s == params[:study_id] }
             else
-              Assay.find(params[:assay_id]).authorized_for('view')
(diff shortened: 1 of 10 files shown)


app/controllers/sops_controller.rb
```diff
@@ -56,6 +56,26 @@ def update
     end
   end
 
+  def dt_typeahead
```
stuzart, line 59: give a more descriptive name, e.g dynamic_table_typeahead, so that it is clearer in the future

test/functional/sops_controller_test.rb
```diff
@@ -2140,6 +2140,52 @@ def test_editing_doesnt_change_contributor
     assert_equal 0, policy.permissions.count
   end
 
+  test 'should return sops only linked to studies or assays' do
```
stuzart, line 2143: what should return sops only? mention the action in the test name

stuzart (changes requested):

stuzart (approved):

Issue #2122, opened 2025-01-27
No observation unit in dashboard overview


jjkoehorst:
**Describe the bug**
When going to the dashboard there is no support for observation unit?

**To Reproduce**
1. Go to dashboard (after creating an observation unit)

**Expected behavior**
It should be there

**Screenshots**
<img width="1040" alt="Image" src="https://github.com/user-attachments/assets/6b4afe4d-c753-4fc2-85a2-4022b78b0ae9" />

PR #2126, opened 2025-01-29
Include observation units in the statistics dashboards

stuzart:
added to the standard, project and programme dashboards

* fix for #2122


fbacall (approved):

stuzart:
> Could also add a colour for them:
> 
> https://github.com/seek4science/seek/blob/698ad57140a5b1e4bf110fb694c5726a290629b1/app/helpers/isa_helper.rb#L16-L33

Thanks. A blue ecolour is applied and I assumed it was picking this up from a colour I'd configured for the ISA graph previously. But looks like I'd done neither and blue is just the default. I will update this before merging.

PR #2133, opened 2025-02-03
fix missing test assertions

stuzart:
Updated tests that were reporting missing assertions


fbacall (changes requested):
One slipped through: https://github.com/seek4science/seek/actions/runs/13076419889/job/36489622224#step:15:13

```
Test is missing assertions: `test_subclasses_must_implement_ontology_file_method` /home/runner/work/seek/seek/test/unit/ontologies/ontology_reader_test.rb:83
```

otherwise looks good

fbacall (approved):

PR #2135, opened 2025-02-04
Add ROR Support for Institutions

whomingbird:
ROR ([Research Organization Registry](https://ror.org/)) is an open, community-driven registry of research organizations worldwide. It provides a unique, persistent identifier (ROR ID) for institutions, making it easier to disambiguate organizations in research data. ROR IDs are openly available, interoperable with other identifier systems (like ORCID and CrossRef), and designed for integration into research workflows.

This PR introduces ROR support in the institutions module. Users can enter institution data in three ways:

- Search by Institution Name – Type the institution's title and select it from the suggested list.
- Enter a ROR ID – Retrieve institution data by entering its ROR ID.
- Manual Entry – If the institution does not have a ROR ID, manually fill in the required fields.

The metadata for options 1 and 2 is fetched directly from ROR's API.

Additionally, ROR integration has been extended to project creation, enabling users to associate projects with institutions—either from locally stored data or dynamically fetched from ROR. 

**Improvements in Details** 

- Database
Added a new column `ror_id` to the institutions table.
Ensured `ror_id` is unique for institutions.

- UI
Added the new input field of ROR ID 
The ROR ID is displayed as a clickable link in the institution details.
Added a "clear fields" button when creating institutions.
Disabled input fields when 
       metadata is loaded either from locally stored data 
       or dynamically fetched from ROR. 
       or editing an institution with an existing ROR ID.
Improved CSS and JavaScript for better responsiveness.

- Functionality

Users can type an institution name and see a suggested list with ROR data.
ROR ID validation implemented to prevent incorrect entries.
Improved error messages when an invalid ROR ID is entered.
Users can fetch institution metadata via ROR ID.
When creating an institution along with a project, the ROR ID is now saved properly.
Converted country codes to full country names for better readability.

- API

Added `ror_id` to the Institution API (Read/Write).
Updated API examples to include the new `ror_id` attribute.


app/views/institutions/_select_or_define.html.erb
```diff
@@ -5,77 +5,48 @@
...
 
-  <label class="control-label">Type the name of the <%= t('institution') %> </label><span class="required">*</span>
-
-  <%# I am setting the institution id to a junk value of 0 (if blank) to get around a bug in select2 wherein an <option>
-      with a blank `value`, that is selected when the page is loaded, does not get rendered as a "tag". %>
-  <%= objects_input('institution[id]', @institution.title.present? ? [@institution.tap { |i| i.id ||= 0 }] : [],
-                    typeahead: { query_url: typeahead_institutions_path,
-                                 handlebars_template:'typeahead/institution' },
-                    limit: 1, allow_new: true, class: 'form-control') -%>
+  <label class="control-label">Type the name of the <%= t('institution'.downcase) %> </label><span class="required">*</span>
```
stuzart, line 25: `.downcase` should be outside the bracket

stuzart (changes requested):

app/views/institutions/_institution_fields.html.erb
```diff
@@ -0,0 +1,81 @@
...
+    </small>
+    <button type="button" id="fetch-ror-data-with-id" class="btn btn-sm btn-primary">fetch</button>
+  </div>
+        <div class="form-group">
+          <%= f.label :city -%>
+          <%= f.text_field :city, :class=>"form-control" -%>
+        </div>
+        <div class="form-group">
+          <%= f.label :country -%>
+          <%= f.text_field :country, value: @institution.nil? ? nil:CountryCodes.country(@institution.country), :class=>"form-control" -%>
```
stuzart, line 45: wouldn't this be better using `f.country_select` to make it safter to select a valid country and country-code rather than typing it in?

stuzart (changes requested):

app/assets/stylesheets/ror-widget.css
```diff
@@ -0,0 +1,31 @@
...
+    box-shadow: 0 5px 10px rgba(0,0,0,.2);
+
+
+}
+
+.league-name {
+    margin: 0 20px 5px 0px;
+    padding: 3px 6px;
+    background-color: rgba(177, 183, 187, 0.47);
+    transition: background-color 0.5s ease-in-out;
```
fbacall, line 29: Try and use bootstrap sass variables for the colours, otherwise it will look weird on instances that are using a different theme, like we do here: 
https://github.com/seek4science/seek/blob/main/app/assets/stylesheets/styles.scss#L440-L443
e.g. you could use `$input-bg` for the `background-color`
https://github.com/twbs/bootstrap-sass/blob/v3.4.3/assets/stylesheets/bootstrap/_variables.scss#L189-L192

fbacall (changes requested):
I think when creating an institution with a ROR ID, SEEK should fetch the metadata from the ROR API on the backend (in a `before_validation` or something), instead of everything being passed by the client (similar to what we do for publications with pubmed IDs/DOI).

This would do several things:
1. Validates that the ROR ID is real and resolvable.
2. Ensures the metadata has not been modified.
3. Makes it easier to create institutions with ROR IDs via the API.

app/assets/javascripts/institution-ror-typeahead.js
```diff
@@ -0,0 +1,266 @@
...
+    const match = rorUrl.match(regex);
+    if (match) {
+        return match[1];
+    } else {
+        return null;
+    }
+}
+
+function fetchRorData(rorId) {
+    var url = ROR_API_URL + '/' + rorId;
```
stuzart, line 32: is it safe to do this, without explicity asking a users consent? Besides possible cross-origin issues, doesn't this introduce the same issue of passing the IP address to another server, similar to what led to preventing us using google fonts, and embedded youtube videos ?

https://www.cookieyes.com/documentation/google-fonts-and-gdpr/

stuzart (changes requested):

fbacall (commented):
I still think you should fetch all the metadata from ROR on the backend if a `ror_id` is present (especially since you are calling out to it to validate anyway).

for example:
```ruby
before_validation :fetch_from_ror, if: -> { ror_id.present? && ror_id_changed? }

def fetch_from_ror
  response = Ror::Client.new.fetch_by_id(ror_id)
  if response[:error]
    errors.add(:ror_id, response[:error]) 
  else
    # ... assign attributes based on ROR response
  end
end
``` 

This would allow easy creation of institutions by just passing `Institution.new(ror_id: "123abc")`

stuzart (approved):

app/models/institution.rb
```diff
@@ -5,10 +5,16 @@ class Institution < ApplicationRecord
...
+  before_validation :fetch_ror_details, if: -> { ror_id.present? && ror_id_changed? }
+
   validates :title, uniqueness: true
+  validates :ror_id, uniqueness: true, allow_blank: true
   validates :web_page, url: { allow_nil: true, allow_blank: true }
   validates :country, country: true
 
+
+  validates :ror_id, uniqueness: true, allow_blank: true
+
```
fbacall, line 17: This is duplicated

fbacall (approved):
Just one minor comment, but otherwise looks good

PR #2161, opened 2025-03-05
Organise Gemfile and update all gems

fbacall:
- Refactors the Gemfile to:
  - Unlock all gem versions (except Rails) unless there is an explicit reason (provided as a comment)
  - Sort gems alphabetically
  - Remove some unused gems
- Updates to latest possible versions of various gems (via `bundle update`), and fixes code broken by those updates (mostly commonmarker-related)
- Removes unused `strain_descendants` table.


stuzart (changes requested):
just a question about the recaptcha setting

stuzart (approved):

PR #2168, opened 2025-03-12
Fix registered sop issue

kdp-cloud:
Minor fix for the dynamic table

### DIFF
--- a/app/controllers/sops_controller.rb
+++ b/app/controllers/sops_controller.rb
@@ -57,13 +57,15 @@ class SopsController < ApplicationController
   end
 
   def dynamic_table_typeahead
-    return if params[:study_id].blank? && params[:assay_id].blank?
+    study_id = params[:study_id] if params[:study_id].present? && !%w[null undefined].include?(params[:study_id].to_s)
+    assay_id = params[:assay_id] if params[:assay_id].present? && !%w[null undefined].include?(params[:assay_id].to_s)
+    return if study_id.blank? && assay_id.blank?
 
     query = params[:query] || ''
-    asset = if params[:study_id].present?
-              Study.authorized_for('view').detect { |study| study.id.to_s == params[:study_id] }
+    asset = if study_id.present?
+              Study.authorized_for('view').detect { |study| study.id.to_s == study_id.to_s }
             else
-              Assay.authorized_for('view').detect { |assay| assay.id.to_s == params[:assay_id] }
+              Assay.authorized_for('view').detect { |assay| assay.id.to_s == assay_id.to_s }
             end
 
     sops = asset&.sops || []
--- a/app/controllers/sops_controller.rb
+++ b/app/controllers/sops_controller.rb
@@ -66,14 +66,14 @@ class SopsController < ApplicationController
 
     query = params[:query] || ''
     asset = if study_id.present?
-              study = Study.includes(:sops).find(study_id)
+              study = Study.includes(:sops).find_by_id(study_id)
               study if study&.can_view?
             else
-              assay = Assay.includes(:sops).find(assay_id)
+              assay = Assay.includes(:sops).find_by_id(assay_id)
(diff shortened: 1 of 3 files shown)


app/controllers/sops_controller.rb
```diff
@@ -57,13 +57,15 @@ def update
...
-    return if params[:study_id].blank? && params[:assay_id].blank?
+    study_id = params[:study_id] if params[:study_id].present? && !%w[null undefined].include?(params[:study_id].to_s)
+    assay_id = params[:assay_id] if params[:assay_id].present? && !%w[null undefined].include?(params[:assay_id].to_s)
+    return if study_id.blank? && assay_id.blank?
 
     query = params[:query] || ''
-    asset = if params[:study_id].present?
-              Study.authorized_for('view').detect { |study| study.id.to_s == params[:study_id] }
+    asset = if study_id.present?
+              Study.authorized_for('view').detect { |study| study.id.to_s == study_id.to_s }
```
stuzart, line 66: looks quite heavyweight to get all the visible studies and then iterate to get the one with the id. Why not `Study.find_by_id(id)&.can_view?` and then set `asset` to `nil` if its hidden or doesn't exist ?

app/controllers/sops_controller.rb
```diff
@@ -57,13 +57,15 @@ def update
...
             else
-              Assay.authorized_for('view').detect { |assay| assay.id.to_s == params[:assay_id] }
+              Assay.authorized_for('view').detect { |assay| assay.id.to_s == assay_id.to_s }
```
stuzart, line 68: same as mentioned for Study

test/functional/sops_controller_test.rb
```diff
@@ -2184,6 +2184,21 @@ def test_editing_doesnt_change_contributor
...
+    # Should return 5 sops linked to assay: nr. 16 to 20 like before
+    get :dynamic_table_typeahead, params: { study_id: 'undefined', assay_id: assay.id, query: 'assay' }, format: :json
+    results = JSON.parse(response.body)['results']
+    assert_equal results.count, 5
+
+
+    # Query '1'
+    # assay_id is 'null'
+    # Should return 2 sops linked to study: nr. 1 and 10 like before
+    get :dynamic_table_typeahead, params: { study_id: study.id, assay_id: 'null', query: '1' }, format: :json
```
stuzart, line 2199: because study_id is present and checked first, it doesn't matter what value for assay_id is passed, I think `assay_id: 'fred'` would give the same result

stuzart (changes requested):

app/controllers/sops_controller.rb
```diff
@@ -57,22 +57,35 @@ def update
...
+
+    if study_id.blank? && assay_id.blank?
+      raise "Invalid parameters! Either study id '#{params[:study_id]}' or assay id '#{params[:assay_id]}' must be a valid id."
+    end
 
     query = params[:query] || ''
-    asset = if params[:study_id].present?
-              Study.authorized_for('view').detect { |study| study.id.to_s == params[:study_id] }
+    asset = if study_id.present?
+              study = Study.includes(:sops).find(study_id)
```
stuzart, line 69: using `.find` will raise an error if the study_id doesn't exist, whereas `find_by_id` will return nil. Using find_by_id will more closely match the previous behaviour.

app/controllers/sops_controller.rb
```diff
@@ -57,22 +57,35 @@ def update
...
+              study if study&.can_view?
             else
-              Assay.authorized_for('view').detect { |assay| assay.id.to_s == params[:assay_id] }
+              assay = Assay.includes(:sops).find(assay_id)
```
stuzart, line 72: same comment about using `find_by_id` applies here.

stuzart (approved):
Approved, but added a comment you might want to concider before merging.

Issue #2134, opened 2025-02-03
Generate Extended Metadata and Sample Types from a fairdata station TTL file


stuzart:
Parse the file to extract the relevant information, and then from that generate the JSON for EMT's registration. 
Show a preview of the EMT's and an option to change the title, before submitting

- [x] Parse TTL file, and get the attributes, pid, and labels
- [x] Create extended metadata JSON for each type per I.S.O.A, with default titles for each type
- [x] Show preview of each JSON definition (in a way that can be reused for general EMT JSON submission)
- [x] Submit to create
- [x] Allow titles to be modified
- [x] Extend to also preview and create the Sample Type
- [x] Update to automatically detect for duplicates - where all attributes match
- [x] Merge the properties across multiple instances of the same type
- [x] Ensure features only shown if fair-data-station is enabled
- [x] Don't match with sample types if auth says they are hidden

PR #2176, opened 2025-03-18
EMT and Sample Type creation from FD-Station

stuzart:
Creation of Extended Metadata Types and Sample Type from the information contained in a FAIR Data Station Turtle output

* Fix for #2134


lib/seek/fair_data_station/annotation_details.rb
```diff
@@ -0,0 +1,20 @@
...
+        @property_id = attributes[:property_id]
+        @label = attributes[:label]
+        @description = attributes[:description]
+        @pattern = attributes[:pattern]
+        @required = attributes[:required]
+      end
+
+      def to_extended_metadata_attribute_json
+        { 'title' => label, 'description' => description, 'pid' => property_id, 'required' => required,
+          'type' => 'String' }
```
whomingbird, line 16: is only the type "String" supported ?

app/controllers/application_controller.rb
```diff
@@ -385,7 +385,7 @@ def log_event
         end
       when *Seek::Util.authorized_types.map { |t|
  t.name.underscore.pluralize.split('/').last } + ["sample_types"] # TODO: Find a nicer way of doing this...
-        action = 'create' if action == 'create_metadata' || action == 'create_from_template'
+        action = 'create' if action == 'create_metadata' || action == 'create_from_template'|| action == 'create_from_fair_ds_ttl'
```
whomingbird, line 388: log event is not working, also there is no tests for this

whomingbird:
The code looks clear and well-written; only one place needs to check again

app/controllers/application_controller.rb
stuzart: it was working and tested for Sample Type, but I've updated it for EMT's too along with tests

PR #2180, opened 2025-03-23
Add Morpheus Button for Loading and Simulating Public Models

whomingbird:
[Morpheus](https://morpheus.gitlab.io/) is a modeling and simulation environment for the study of multi-scale and multicellular systems.

This PR introduces a Morpheus button that allows users to load publicly available Morpheus models into their locally installed Morpheus application and start simulations. The feature is similar to the existing Copasi button.

Key Features:

- System administrators can enable or disable the Morpheus button.
- Currently supports only Morpheus models.
- Works exclusively with publicly downloadable models.


lib/seek/config_setting_attributes.yml
```diff
@@ -57,6 +57,7 @@ jerm_enabled:
 email_enabled:
 jws_enabled:
 copasi_enabled:
+morpheus_enabled:
```
stuzart, line 60: needs a default setting in `config/initializers/seek_configuration.rb` - which I think should be true

test/functional/morpheus_test.rb
```diff
@@ -0,0 +1,48 @@
...
+      "morpheus://#{request.host_with_port}/models/#{model.id}/content_blobs/#{blob.id}/download"
+    end
+
+    public_model = FactoryBot.create(:morpheus_model, policy: FactoryBot.create(:public_policy))
+    teusink_model = FactoryBot.create(:teusink_jws_model, policy: FactoryBot.create(:public_policy))
+    private_model = FactoryBot.create(:morpheus_model, policy: FactoryBot.create(:private_policy))
+
+    # for a public Morpheus model, if Morpheus is not enabled, the button is hidden
+    with_config_value(:morpheus_enabled, false) do
+      get :show, params: { id: public_model }
```
stuzart, line 20: would be good to have each of these `get :show` requests and test in separate tests, to avoid problems when `rack` is updated to v3+

stuzart (changes requested):
looks good, just a couple of minor changes

lib/seek/config_setting_attributes.yml
whomingbird: shoud i also add a default setting in config/initializers/seek_configuration.rb  for copasi, 
Seek::Config.default :copasi_enabled, true

PR #2194, opened 2025-05-01
WorkflowHub fixes

fbacall:
- Fixes issue parsing RO-Crate metadata with authors that have affiliation as an array
- Fixes issue where old files would not be cleared out when submitting new version as an RO-Crate


lib/seek/workflow_extractors/base.rb
```diff
@@ -78,8 +78,12 @@ def extract_author(obj)
             if affiliation.is_a?(String)
               author[:affiliation] = affiliation
             else
-              affiliation = affiliation.dereference if affiliation.respond_to?(:dereference)
-              author[:affiliation] = affiliation['name'] if affiliation && affiliation['name'].present?
+              affiliation = [affiliation] unless affiliation.is_a?(Array)
```
stuzart, line 81: you can actually just do `affiliation = Array(affiliation)` to acheive the same thing, but I don't think it matters

PR #2231, opened 2025-06-26
Discipline annotations

fbacall:
- Allows projects and workflows to be annotated with scientific disciplines.
- Propagates disciplines set on a project down to workflows under that project (if they do not already have any disciplines set).
- Refactors how default policies/licenses are handled, fetching them with an AJAX call rather than storing a mapping on the page.
- Moves investigation/study select field to the "manage" pages of studies/assays.

### DIFF
--- a/app/views/projects/_form.html.erb
+++ b/app/views/projects/_form.html.erb
@@ -61,8 +61,6 @@
     <%= render :partial=> "projects/select_add_current_user_to_project" %>
 <% end %>
 
-<%= render partial: "assets/controlled_vocab_annotations_form_property", locals: { resource: @project, property: :disciplines} -%>
-
 <%= f.fancy_multiselect :organisms, { possibilities: Organism.order(:title).all, hidden: false } %>
 
 <%= f.fancy_multiselect :human_diseases, { possibilities: HumanDisease.order(:title).all, with_new_link: true, hidden: false } %>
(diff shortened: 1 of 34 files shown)


app/views/workflows/provide_metadata.html.erb
```diff
@@ -61,6 +61,8 @@
 
     <%= render partial: 'assets/controlled_vocab_annotations_form_properties', locals: { resource: @workflow } %>
 
+    <%= render partial: 'assets/controlled_vocab_annotations_form_property', locals: { resource: @workflow, property: :disciplines} -%>
```
stuzart, line 64: could this be included in `assets/controlled_vocab_annotations_form_properties` for consistency and so they are always grouped together

app/views/projects/_form.html.erb
```diff
@@ -62,6 +61,8 @@
     <%= render :partial=> "projects/select_add_current_user_to_project" %>
 <% end %>
 
+<%= render partial: "assets/controlled_vocab_annotations_form_property", locals: { resource: @project, property: :disciplines} -%>
```
stuzart, line 64: I think it would be better if this was placed next to Topics, since the are related, or better replace both with an updated  `assets/controlled_vocab_annotations_form_properties`

stuzart (changes requested):
a couple of minor comments, but also I do wonder if we now need Topics anymore, as this is just a duplication with a smaller set of terms. Alternatively the extra terms could be added to Topics. Having both seems a bit confusing and weird.

stuzart (approved):

Issue #2230, opened 2025-06-25
Don't send error emails for Errno::ENOENT exceptions


stuzart:
All errors receieved for this Exception in recent history related to the `  No such file or directory @ rb_sysopen ` when reading the settings cache, which can be safely ignored - so they just clutter up emails and hide more important errors

PR #2251, opened 2025-06-30
rescue an Errno::ENOENT error when thrown by the settings cache 

stuzart:
when there is an error it will revert to reading the setting directly

* fix for #2230


fbacall (approved):
I think the `No such file or directory @ rb_sysopen ... .permissions_check.` errors are fixed in Rails 7+, but there is a similar error we still get:
```
Errno::ENOENT
No such file or directory @ rb_file_s_rename - (/xxx/tmp/cache/settings-cache/.seek_config20250601-1234567-a12345bl, /xxx/tmp/cache/settings-cache/123/ABC/seek_config)
```
that this should handle

PR #2254, opened 2025-07-02
Add sample attribute type field to the cumulative table

### LINKED ISSUE #2235: Fix non-text attributes in experiment view
The experiment view table shows non-text attributes as `[object Object]`.

Image

It should show the title in a badge instead.

kdp-cloud:
fixes #2235 
image


stuzart (commented):
I'm not entirely clear how this works.

Is `s.sample_attribute_type` somehow automatically being turned into a JSON representation (where?), which is then being handled in some pre-existing javascript to display according to the type ?

Issue #2227, opened 2025-06-24
Clean up FairDataStationUpload's after an investigation or project is destroyed


stuzart:
needs appropriate associations and dependant destroy, and also check the content blob and policies are cleaned up

PR #2262, opened 2025-07-09
Clean up FAIR Data Station Uploads

stuzart:
Delete associated `FairDataStationUploads ` when an Investigation is deleted - setting up the appropriate `dependent: :destroy` to also delete the policy and mark the associated content blob for deletion.

Decided against doing the same for Project, as the Investigation could be associated to multiple projects - and for the Project to be deleted the associated items need to be moved or deleted, making it unnecessary. However, the maintenance task will clean up failed imports after 4 weeks (after which time it can be investigated). 

* fix for #2227


whomingbird (approved):
Looks good

PR #2270, opened 2025-07-15
1798 add isa compliant as criterial for filtering results when browsing

### LINKED ISSUE #1798: Add ISA-compliant as criterial for filtering results when browsing
Add ISA-compliant as criterial for filtering results when browsing ISA levels
- ISA-JSON compliant
- Not ISA-JSON compliant

kdp-cloud:
Adds an 'ISA-JSON compliance' filter for:
- Studies
- Assays
- Investigations

Resolves #1798


fbacall (changes requested):
I actually think you could simplify this a lot.
1. Your `BooleanFilter` could be a subclass of `Filter`, and you would just need to call `super` with your own `value_mapping` (e.g. to turn `["true"]` into the boolean `true`) and `label_mapping` (to turn the `true`/`false` booleans into the strings `"Yes"`/`"No"`)
2. You could create a partial to render the boolean filter `app/views/filtering/_boolean_filter.html.erb` to ensure both Yes/No options are there (options are excluded by default if they have 0 results), and to set the `replace: true` flag on the `filter_link`s

PR #2305, opened 2025-08-28
Different rdf formats

jjkoehorst:
When entering a page with the .rdf extension it provides it in the more nicer reading TURTLE format but this is actually syntactically incorrect?... So I added an option such that if you perform .rdf it gives you the RDF/XML format and if you access the page with .ttl it gives you the TURTLE format. 

Example is only for projects at the moment. if needed I can push it for all pages...


jjkoehorst:
As I was working in a branch of a branch I might have committed more than I bargained for...

PR #2405, opened 2025-11-14
Update docker images

kdp-cloud:
- Update the seek images to the latest debian release: 13-trixie
- Multi-stage building of the image, reducing the attack vector and the image size (<2GB)
- Use MySQL 8.4 instead of 8.0
- Patch python dependencies
- Simplify docker compose file by introducing shared service declarations
- Add health checks in docker compose file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,7 +82,7 @@ RUN mkdir -p /var/www
 
 # Fix permissions
 RUN chown www-data:www-data config/initializers public sqlite3-db /var/www
-RUN chown -R www-data:www-data public/api
+RUN chown -R www-data:www-data config public/api
 RUN chmod -R 755 docker/upgrade.sh docker/start_workers.sh
 
 # Python dependencies from requirements.txt


Dockerfile
```diff
@@ -1,82 +1,123 @@
...
+WORKDIR $APP_DIR
+
+# Disable ssl from the mysql client
+RUN echo "[client]\nskip-ssl" > /etc/mysql/conf.d/disable-ssl.cnf
+
+FROM base AS builder
+
+# Install build dependencies
+RUN apt-get update -qq && \
+    apt-get install -y --no-install-recommends build-essential cmake \
```
stuzart: also keep alphabetical. Could these all be moved to the base? Would easier to maintain if all in one place

fbacall (changes requested):
I did a global search for `File.write` and found some cases where SEEK will write to `config/`, so either those writes should be moved somewhere else, or it needs permission to write there.

(And quite possibly there are other places that write to the disk without explicitly using `File.write`)

kdp-cloud:
> > I did a global search for `File.write` and found some cases where SEEK will write to `config/`, so either those writes should be moved somewhere else, or it needs permission to write there.
> > (And quite possibly there are other places that write to the disk without explicitly using `File.write`)
> 
> config needs to be writable for a couple of cases during startup where it updates the config files for Mysql or for whether search is enabled: https://github.com/seek4science/seek/blob/main/docker/shared_functions.sh#L27
> 
> https://github.com/seek4science/seek/blob/main/docker/shared_functions.sh#L50
> 
> Also looks like seek.crontab needs to be writable: https://github.com/seek4science/seek/blob/main/docker/shared_functions.sh#L66

The config folder is now writable again.

fbacall (changes requested):
See Stuart's comment about `seek.crontab`

stuzart (approved):
I will build and deploy on the testing server after 1.17.2

PR #2408, opened 2025-11-17
Extend publication type list

whomingbird:
* fix for #2409

### DIFF
--- a/app/models/publication_type.rb
+++ b/app/models/publication_type.rb
@@ -181,13 +181,4 @@ class PublicationType < ActiveRecord::Base
     other_type&.id
   end
 
-  begin
-    type_registry.each_key do |title|
-      define_singleton_method(title.gsub(/\s+/, '')) do
-        find_by(key: type_registry[title]['key'])
-      end
-    end
-  rescue => e
-    Rails.logger.error("Failed to load publication types: #{e.message}")
-  end
 end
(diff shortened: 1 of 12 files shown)


app/models/publication_type.rb
```diff
@@ -1,143 +1,58 @@
...
+    'phdthesis' => 'phdthesis',
+    'mastersthesis' => 'mastersthesis',
+    'bachelorsthesis' => 'bachelorsthesis'
+  }.freeze
+
+
+  # Load expected publication types from YAML
+  def self.type_registry
+    @type_registry ||= begin
+                         path = Rails.root.join('config/default_data/publication_types.yml')
```
stuzart: don't use the yaml file, which is used for seeding. Fetch this info from the database where we know it is up to date

config/default_data/publication_types.yml
```diff
@@ -0,0 +1,153 @@
...
+  title: Bachelor's Thesis
+  key: bachelorsthesis
+
+MastersThesis:
+  title: Master's Thesis
+  key: mastersthesis
+
+DiplomaThesis:
+  title: Diplom Thesis
+  key: diplomthesis
```
stuzart: should be Diplom**a** thesis

app/models/publication_type.rb
```diff
@@ -1,143 +1,58 @@
...
+    datacite_key = BIBTEX_TO_DATACITE_KEY[publication_key] || 'other'
+    pub_type = PublicationType.find_by(key: datacite_key)
+    return pub_type.id if pub_type
+    other_type = PublicationType.find_by(key: 'other')
+    other_type&.id
+  end
+
+  begin
+    type_registry.each_key do |title|
+      define_singleton_method(title.gsub(/\s+/, '')) do
```
stuzart: are these methods,  e.g `Publication.Journal` or `Publication.Collection` to get the type, ever actually used anywhere?

stuzart (changes requested):

app/models/publication_type.rb
whomingbird: removed

stuzart (approved):

PR #2424, opened 2025-11-26
should show publication and SEEK ID only once when fulltext enabled

whomingbird:
Issue 2420


stuzart (changes requested):
just a minor comment

stuzart (approved):

PR #2427, opened 2025-11-27
Small fixes for managing isa-json-compliant resources

### LINKED ISSUE #2425: Inheriting permissions from assay stream breaks assay permissions
When clicking the 'apply permissions to child assays', it breaks the permissions of those assays.

Image

This error is returned when viewing the project:

```
An ActionView::Template::Error occurred in single_pages#show:

  undefined method `access_type' for nil
  app/helpers/resource_list_item_helper.rb:219:in `list_item_visibility'
```

**Edit:** It appears that Sample Type permissions are not updated with the the assay permissions

### LINKED ISSUE #2428: Wrong redirection from manage page when canceling
I would be nicer to return to the exact page in question when canceling the 'managing' of a resource.
Example: When clicking cancel on the manage study page, it would be nice if the user returns to the single page view if the study is isa-json compliant.

### LINKED ISSUE #2429: Hide the 'propagate permissions checkbox' when not needed
The assay stream's 'propagate permissions to child assays' checkbox is now currently showing all the time, even at creation or when the the assay stream does not hold any child assays. This checkbox should be hidden in this case.

kdp-cloud:
- Propagate assay stream permissions to assay AND assay sample type.
  - Fixes #2425 
- Correct redirections form the manage page of isa-json compliant studies and assay.
  - Fixes #2428
- Show 'propagate permissions' checkbox only when the assay stream has child assays.
  - Fixes #2429

### DIFF
--- a/app/controllers/assays_controller.rb
+++ b/app/controllers/assays_controller.rb
@@ -187,7 +187,7 @@ class AssaysController < ApplicationController
       # Clone the policy from the parent assay
       assay_policy = @assay.policy.deep_copy
       assay.update(policy: assay_policy)
-      assay.sample_type.update(policy: assay_policy)
+      assay.sample_type.update(policy: assay_policy) unless assay.sample_type.nil?
       current_assay_policy.destroy if current_assay_policy
       update_sharing_policies assay
     end
--- a/app/views/assets/_manage_form.html.erb
+++ b/app/views/assets/_manage_form.html.erb
@@ -43,7 +43,7 @@
   <%= render partial: 'assets/manage_specific_attributes', locals: {f: f, sharing_link: sharing_link} %>
 
   <%
-    if @isa_json_compliant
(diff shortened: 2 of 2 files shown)


app/controllers/assays_controller.rb
```diff
@@ -182,7 +185,9 @@ def propagate_permissions_to_children
 
       current_assay_policy = assay.policy
       # Clone the policy from the parent assay
-      assay.update(policy: @assay.policy.deep_copy)
+      assay_policy = @assay.policy.deep_copy
+      assay.update(policy: assay_policy)
+      assay.sample_type.update(policy: assay_policy) unless assay.sample_type.nil?
```
stuzart: bear in mind that this shares the policy between the assay and sample-type - so if the permissions for the assay change, so will with sample type. I can see why this may be intentional

PR #2448, opened 2025-12-10
Expose isa tags through api

kdp-cloud:
Enables users to fetch ISA tags through the API

### DIFF
--- a/test/integration/sitemap_test.rb
+++ b/test/integration/sitemap_test.rb
@@ -71,8 +71,10 @@ class SitemapTest < ActionDispatch::IntegrationTest
     doc = Nokogiri::XML.parse(response.body)
     doc.remove_namespaces!
 
-    # +1 for root path, and -1 for disabled SOPs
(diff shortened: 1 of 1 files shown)


app/serializers/isa_tag_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ISATagSerializer < ActiveModel::Serializer
```
stuzart: I think this needs to be based off `BaseSerializer` or `SimpleBaseSerializer` to get the jsonapi and meta blocks

stuzart (changes requested):
I'm not sure it's JSONAPI compliant and consistent with the rest of the api.
It needs adding to the schema and docs, with examples, which would make this clearer: e.g https://testing.sysmo-db.org/api#tag/sampleAttributeTypes

Ideally also included in `integration/api` tests (although I realise sampleAttributeTypes may not have)

PR #2473, opened 2026-01-15
Introduce Input ISA tag

kdp-cloud:
- Add Input ISA tag


lib/tasks/seek_upgrades.rake
```diff
@@ -20,6 +20,9 @@ namespace :seek do
     update_morpheus_model
     db:seed:018_discipline_vocab
     strip_publication_abstracts
+    db:seed:015_isa_tags
```
stuzart: does 017_minimal_starter_isa_templates.seeds.rb‎ also need adding here?

Issue #2413, opened 2025-11-18
Event types, and also hybrid/offline option


stuzart:
Add a selection for Event type - current candidates are deliverable, milestone, experiment, experimental phase, presentation, meeting but check for a CV list that could be reused. The fitering should also support this.
Also have a option to say whether it is online/hybrid or in person.

PR #2479, opened 2026-01-22
Event type and Event location

stuzart:
* fix for #2413


lib/tasks/seek_upgrades.rake
```diff
@@ -24,6 +24,7 @@ namespace :seek do
     assign_isa_tag_id_to_template_attributes
     db:seed:017_minimal_starter_isa_templates
     db:seed:019_sop_type_controlled_vocab
+    db:seed:019_event_types
```
kdp-cloud: Nitpicking: Duplicate number of the `019_sop_type_controlled_vocab` seed

lib/seek/filterer.rb
```diff
@@ -73,6 +73,11 @@ class Filterer
             value_field: 'country',
             label_mapping: MAPPINGS[:country_name]
         ),
+        event_type: Seek::Filtering::Filter.new(
```
kdp-cloud, line 76: Question: Does it need unit tests?

kdp-cloud (approved):
Just a couple of minor comments... But nothing worth blocking.

PR #2500, opened 2026-02-18
2478 implement sop type in json export

kdp-cloud:
Adds the protocol type to the ISA-JSON export


lib/isa_exporter.rb
```diff
@@ -322,11 +322,24 @@ def convert_protocol(sop, id, protocol_attrbute, parameter_attributes)
...
+      # else it will revert to the 'old' way of dealing with protocol types
+      sop_type = sop.sop_type_annotation_values&.first if sop.is_a? Sop
+      if sop_type
+        isa_protocol[:protocolType] = {
+          annotationValue: sop_type&.label,
+          termAccession: sop_type&.iri,
+          termSource: sop_type&.sample_controlled_vocab&.source_ontology,
+        }
+      else
+        isa_protocol[:protocolType] = {
```
stuzart, line 336: `if sop.is_a? Sop` - if not, I think it will enter this block as sop_type will be nil. Maybe have the check as a guard at the top of the method ?

kdp-cloud: That is the expected behavior. If sop_type is `nil`, the `protocol_type` attribute should get the values from the sample attribute labeled with the `protocol` isa tag. Which is the 'old' way of populating `protocol_type` and still have to support.

For this reason I can't put it as a guard clause at the top.

Issue #2286, opened 2025-07-29
Block local file upload by users


svbeuningen:
As an admin I want to block local file upload and allow (force) only Remote URLs. When using remote URLs we also want to block the "fetch a copy"  option. Is this option already available in the current config? If not, would it be possible to add this option?

We intent to use FAIRDOM-SEEK only as a catalog of items that are stored in different locations (like S3) and we want to prevent local file storage in FAIRDOM-SEEK.

We also noticed that FAIRDOM-SEEK puts a symlink on top of the Remote URL and the remote URL is not directly visible. Is this also an feature? Ideally a user can show the Remote URL directly so it knows where the data file is stored.

Hopefully someone can help us out.

Best,
Sam (UMC Utrecht, The Netherlands)

PR #2534, opened 2026-03-26
option to block local file updates

stuzart:
* fix for #2286 

Also allows controlling if a link is shown, or downloaded via SEEK. The setting already existed but needed adding to the admin page.

Doesn't eliminate all uploads, still allowed are avatars, model image, project DMP, Fair data station rdf, sample templates.

### DIFF
--- a/app/controllers/content_blobs_controller.rb
+++ b/app/controllers/content_blobs_controller.rb
@@ -12,6 +12,10 @@ class ContentBlobsController < ApplicationController
   api_actions :show, :update, :download
 
   def update
+    if Seek::Config.block_file_uploads
+      raise Seek::UploadHandling::DataUpload::UploadBlockedException, 'Data upload is not allowed.'
+    end
+
     if @content_blob.no_content?
       @content_blob.tmp_io_object = get_request_payload
       @content_blob.save
@@ -235,6 +239,12 @@ class ContentBlobsController < ApplicationController
       params.values.detect { |v| v.is_a?(ActionDispatch::Http::UploadedFile) }
     else
       request.body
+    end
+  end
+  # overrides from DataUpload module
+  def handle_upload_blocked_exception(exception)
+    respond_to do |format|
+      format.all { render plain: exception.message, status: :forbidden }
     end
   end
 end
--- a/app/views/assets/_upload.html.erb
+++ b/app/views/assets/_upload.html.erb
@@ -43,7 +43,7 @@
     <% end %>
 
     <% unless hide_remote %>
-      <div role="tabpanel" class="tab-pane" data-role="seek-upload-field-tab-pane" data-tab-id="remote-url">
+      <div role="tabpanel" class="tab-pane <%= hide_local ? 'active' : '' %>" data-role="seek-upload-field-tab-pane" data-tab-id="remote-url">
         <div class="form-group">
           <div class="row">
             <div class="col-sm-8">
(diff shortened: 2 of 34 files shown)


test/unit/upload_handling_test.rb
```diff
@@ -225,6 +225,31 @@ class UploadHandingTest < ActiveSupport::TestCase
...
+
+  test 'exception in handle upload if upload data present and blocked' do
+    with_config_value(:block_file_uploads, true) do
+      @params = { content_blobs: [{ data_url: 'http://fish.com', data: fixture_file_upload('empty_file', 'text/plain') }] }
+      assert_raises(Seek::UploadHandling::DataUpload::UploadBlockedException, match: /Data upload is not allowed\. Please provide a URL to the data instead/) do
+        handle_upload_data
+      end
+      @params = { content_blobs: [{ data_url: 'http://fish.com', base64_data: fixture_file_upload('empty_file', 'text/plain') }] }
+      assert_raises(Seek::UploadHandling::DataUpload::UploadBlockedException, match: /Data upload is not allowed\. Please provide a URL to the data instead/) do
+        handle_upload_data
```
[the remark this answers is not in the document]
stuzart: this is incorrect

app/controllers/content_blobs_controller.rb
```diff
@@ -12,6 +12,10 @@ class ContentBlobsController < ApplicationController
   api_actions :show, :update, :download
 
   def update
+    if Seek::Config.block_file_uploads
+      raise Seek::UploadHandling::DataUpload::UploadBlockedException, 'Data upload is not allowed.'
+    end
+
```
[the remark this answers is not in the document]
stuzart: this is incorrect

fbacall (commented):
Should it also block uploads for Workflows? And via the Git interface?

Image

Image

app/views/assets/_upload.html.erb
```diff
@@ -20,86 +21,98 @@
...
+        <% if batch %>
+          <div class="form-group">
+
+            <div>
+              <div class='alert alert-warning' role='seek-url-checker-remind-to-add-file' style='display:none'>
+                Please click on "<b>Add</b>" button to confirm adding this Remote URL to <%= Seek::Config.instance_name %> !</div></div>
+            <%= button_link_to 'Add', 'new', '#', 'data-role' => 'seek-upload-field-add-remote' %>
+          </div>
+          <%= content_tag :script,
+                          json_escape(existing_objects.map { |o| { text: "#{o.url.blank? ? o.original_filename : o.url} (original)", id: o.id } }.to_json),
```
fbacall: This line escapes the JSON and breaks the parsing later on:

```
Uncaught SyntaxError: JSON.parse: expected property name or '}' at line 1 column 3 of the JSON data
```

https://github.com/seek4science/seek/blob/01e0740e2523b47e8c931f048487ac783a9bac28/app/assets/javascripts/upload.js#L10-L12

Noticed when I disabled file uploads, then tried to view a model (with existing files) and attempting to upload a new version.

fbacall: Maybe just needs to `json_escape` the URL/filename?

fbacall: According to this: https://apidock.com/rails/v5.2.3/ERB/Util/json_escape

You just need to call `raw()` on the `json_escape`d string

fbacall (approved):

PR #2535, opened 2026-03-31
Improvements for spreadsheet download

### LINKED ISSUE #2475: Add extended metadata to spreadsheet
Add a new 'extended metadata' metadata  sheet that contains the extended metadata of the associated study or assay.

### LINKED ISSUE #2495: Include Extended metadata in the spreadsheet
It would be nice to be able to include extended metadata in the spreadsheet representation of an ISA-JSON compliant assay or study.

kdp-cloud:
- Split out sheets in partials to improve readability
- Impose at least viewing permission to use data files and strains in spreadsheet
- Add study and assay extended metadata sheet


app/controllers/single_pages_controller.rb
```diff
@@ -82,7 +86,7 @@ def download_samples_excel
 
     notice_message << '</ul>'
     flash[:notice] = notice_message.html_safe
```
stuzart: it may not be html_safe, because of the `study.title` and `assay.title` . These should wrapped with `h(..)` when building the `notice_message`

app/helpers/single_pages_helper.rb
```diff
@@ -59,16 +102,16 @@ def get_values_for_sops(sample_attribute)
...
     end
 
-    sops.map { |sop| { id: sop.id, type: 'Sop', title: sop.title }.to_json }
+    sops.authorized_for(:view).map { |sop| { id: sop.id, type: 'Sop', title: sop.title }.to_json }
   end
 
-  def get_values_for_datafiles(sample_attribute)
-    data_files = sample_attribute.sample_type.projects.first.data_files
+  def get_values_for_datafiles
+    data_files = DataFile.authorized_for(:view)
```
stuzart: I agree with copilot that both for DataFile and Stain, the number would be very large. If it's not possibe to scope it based on projects, would be good to use `in_batches`

stuzart (changes requested):

app/helpers/single_pages_helper.rb
kdp-cloud: I know but this was a 'bug' since this differs from how data files are normally linked to a sample. For data files and strains, the [typeahead](https://github.com/seek4science/seek/blob/87c39e2eab845a19a21428542fb329fe180cd375/lib/seek/assets_common.rb#L45) result is also not restricted to the project.

I am using `find_each` now, which works in batches of 1000 under the hood, in combination with `pluck`. This should address this concern a little bit.

stuzart (approved):

PR #2538, opened 2026-04-05
Seeds improvements

jjkoehorst:
Additional content to provide a more complete test environment.
Added additional elements to various levels within SEEK (e.g. license added, tags added, additional creators)
- Presentation was not working properly first? Should be fixed now
- added a document with a simple text file
- created a collection where many items are bundled together
- gave the admin and guest a first and last name so it does not need to be filled in when testing things locally.


stuzart:
* applied in #2545

PR #2542, opened 2026-04-13
Retract DOI

### LINKED ISSUE #2497: Deleting resources with DOIs
It may be the case that a SEEK resource with a DOI needs to be deleted, e.g. if it contained sensitive data and was published accidentally.

Support being able to delete SEEK resources with a DOI:
- "Hide" the DOI on DataCite
- Remove any content from the SEEK filestore/git
- Blank out metadata (but ensure the resource is still valid)
- Remove from SEEK search index
- Capture the reason for the deletion
- Display a tombstone page if the DOI is visited

Things to consider
- How to implement this with minimal impact on the code
- How to store the info for the tombstone page
- How to stop the resource showing up in "related items" etc., or being counted in statistics (or maybe we don't care about this?)
- Permissions - the user should have manage rights to the resource

https://support.datacite.org/docs/can-i-delete-or-change-my-dois
https://support.datacite.org/docs/tombstone-pages
https://support.datacite.org/docs/updating-metadata-with-the-rest-api#changing-the-doi-state

lasse-cs:
Resolves #2497

Allows a user to send a DELETE DOI request to datacite.

This will mark the resource as no longer findable.

After retracting, an `AssetDoiLog` entry is created to log the retraction and the resource is deleted.

If a user follows the DOI - the AssetDoiLog is queried and a tombstone page is displayed.


app/controllers/application_controller.rb
```diff
@@ -284,7 +284,15 @@ def render_not_found_error(e)
     respond_to do |format|
       format.html do
         User.with_current_user current_user do
-          render template: 'general/landing_page_for_not_found_item', status: :not_found
+          klass = controller_name.singularize.camelize
+          version = params[:version]
+
+          retract_log = AssetDoiLog.where(asset_type: klass, asset_id: params[:id], asset_version: version, action: AssetDoiLog::RETRACT).last
```
fbacall, line 290: I wonder if we should also check for the existence of a `AssetDoiLog` without the `asset_version` too?

For example if there is a DOI on a "parent" resource which gets retracted, and all its versions are thus deleted, but then someone visits e.g. /workflows/123?version=1 somehow.

(No idea if this can happen - disregard if not)

lib/seek/doi/retracting.rb
```diff
@@ -0,0 +1,58 @@
+module Seek
+  module Doi
+    module Retracting
```
fbacall, line 3: I would just put all this code in `Seek::Doi::Minting`, since you can't really have one without the other.

app/models/asset_doi_log.rb
```diff
@@ -7,6 +7,7 @@ class AssetDoiLog < ApplicationRecord
   MINT = 1
   DELETE = 2
   UNPUBLISH = 3
+  RETRACT = 4
```
fbacall, line 10: I've thought a bit about this, and think we should use `DELETE`.

From looking at the two places where `DELETE` and `UNPUBLISH` are used - it seems their intent was to handle cases where something was deleted and something was made private (as in, deleted or unpublished from SEEK, not from the DOI provider), respectively. In our case we're deleting the resource from SEEK, so lets use that.

fbacall (changes requested):
Looks good, some minor comments, the main thing is: lets just use `AssetDoiLog::DELETE` instead of coming up with a new action.

app/views/general/landing_page_for_doi_retracted_item.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="landing_page">
+  <h1>410</h1>
+  <% asset_type = t("#{controller_name.singularize}") %>
+  <h2>The requested <%= asset_type %> does not exist anymore.</h2>
+
+  <p>The <%= asset_type %> with identifier <%= doi_link(retract_log[:doi]) %> has been retracted.</p>
+  <p><%= render_doi_citation(retract_log[:doi], selected_citation_style) %></p>
```
lasse-cs: Not super happy with this at the moment....

I tried https://github.com/datacite/bolognese to get citeproc hash from the metadata to then render the citation, but there were issues with conflicting versions of dependencies.

lasse-cs: I have replaced it with extracting information from the metadata into a `Datacite::Metadata` and then serializing this .

app/views/datacite_doi/retract_doi_confirm.html.erb
```diff
@@ -0,0 +1,36 @@
...
+            <ul>
+                <% @asset.doi_identifiers.each do |doi| -%>
+                    <li>
+                        <pre><%= doi %></pre>
+                    </li>
+                <% end %>
+            </ul>
+
+            <div class="form-group">
+              <%= label_tag :retraction_reason, 'Reason for retraction (optional)' %>
```
fbacall: Lets make this mandatory

app/controllers/application_controller.rb
```diff
@@ -284,7 +284,20 @@ def render_not_found_error(e)
...
+          klass = controller_name.singularize.camelize
+          version = params[:version]
+
+          if version.nil?
+            retract_log = AssetDoiLog.where(asset_type: klass, asset_id: params[:id], action: AssetDoiLog::DELETE).last
+          else
+            retract_log = AssetDoiLog.where(asset_type: klass, asset_id: params[:id], asset_version: version, action: AssetDoiLog::DELETE).last
+          end
+          if retract_log.present?
+            metadata = JSON.parse(retract_log.datacite_metadata) if retract_log.datacite_metadata.present?
```
fbacall: I think we should handle the JSON parsing/serializing in the AssetDoiLog model. You can define some custom getter/setter methods e.g. `metadata`/`metadata=` and do it in there

lib/seek/doi/acts_as_doi_mintable.rb
```diff
@@ -147,6 +174,11 @@ def create_log
           AssetDoiLog.create(asset_type: doi_resource.class.name, asset_id: doi_resource_id, asset_version: doi_resource_suffix,
                              doi: suggested_doi, action: AssetDoiLog::MINT, user_id: User.current_user.try(:id))
         end
+
+        def retract_log(retraction_reason = nil, metadata = nil)
+          AssetDoiLog.create(asset_type: doi_resource.class.name, asset_id: doi_resource_id, asset_version: doi_resource_suffix,
+                             doi: suggested_doi, action: AssetDoiLog::DELETE, user_id: User.current_user.try(:id), comment: retraction_reason, datacite_metadata: metadata)
```
fbacall: You can just assign `user: User.current_user` (the way it was done in `create_log` is also wrong)

lib/seek/doi/acts_as_doi_mintable.rb
```diff
@@ -66,6 +62,33 @@ def mint_doi
...
+          unless Seek::Config.doi_minting_enabled
+            errors.add(:base, 'DOI minting is not enabled')
+            return false
+          end
+
+          endpoint = datacite_client
+          metadata = Datacite::MetadataReader.new(endpoint.metadata(doi)).parse
+          endpoint.inactivate(doi)
+
+          retract_log(retraction_reason, metadata.to_json)
```
fbacall: Validate `retraction_reason` is present

lib/seek/doi/acts_as_doi_mintable.rb
```diff
@@ -109,6 +132,10 @@ def can_mint_doi?
           Seek::Config.doi_minting_enabled && !doi_time_locked? && !has_doi? && visible?(nil)
         end
 
+        def can_retract_doi?
+          Seek::Config.doi_minting_enabled && has_doi? && visible?(nil)
```
fbacall: The `visible?(nil)` check is probably unnecessary

fbacall (changes requested):

fbacall (approved):

Issue #2557, opened 2026-04-30
RDF export: support nested (LinkedExtendedMetadata) attributes as blank nodes


whomingbird:
## Summary

Currently, when an `ExtendedMetadata` attribute has type `LinkedExtendedMetadata` or `LinkedExtendedMetadataMulti`, the RDF export serialises the nested value as a plain string literal (e.g. `"{\"start_date\"=>\"2020-03-01\", \"end_date\"=>\"2030-12-31\"}"`). This produces invalid, unqueryable RDF for any vocabulary that models hierarchical data (e.g. HealthDCAT-AP `health:retentionPeriod`).

## Problem

`extended_metadata_triples` in `lib/seek/rdf/rdf_generation.rb` called `RDF::Literal()` unconditionally on every attribute value. For linked (nested) metadata types this turned the whole nested hash into a single opaque string rather than a structured blank node.

### `lib/seek/rdf/rdf_generation.rb`
- Replaced the flat loop in `extended_metadata_triples` with a recursive `emit_emt_attributes` helper.
- For `LinkedExtendedMetadata` (single): emits `[subject, predicate, blank_node]` then recurses into the nested type's attributes on the blank node.
- For `LinkedExtendedMetadataMulti` (multiple): emits one blank node per item in the array.
- Plain (non-linked) attributes continue to emit `RDF::Literal` as before.
- Only attributes with a `pid` are emitted at each level (nested attributes without a `pid` are silently skipped).

### `test/factories/extended_metadata_types.rb`
Added factories for the RDF nested export tests:
- `rdf_test_nested_period_emt`, `rdf_test_contact_emt`, `rdf_test_partial_pid_nested_emt` — nested EMT shapes
- `rdf_test_data_file_single_nested_emt`, `rdf_test_data_file_multi_nested_emt`, `rdf_test_data_file_partial_pid_emt`, `rdf_test_data_file_flat_emt` — top-level DataFile EMT wrappers with PIDs on linking attributes

### `test/unit/rdf_generation_test.rb`
Four new tests:
1. `LinkedExtendedMetadata` (single) emits a blank node with correct nested triples
2. `LinkedExtendedMetadataMulti` emits one blank node per item
3. Nested attribute without a `pid` is silently skipped
4. Flat string attribute still emits a plain literal (regression guard)

## Example: before / after

**Before**
```
<http://example.org/DataFile/3>
  <http://healthdataportal.eu/ns/health#retentionPeriod>
[...]

PR #2558, opened 2026-04-30
Support nested em rdf export

whomingbird:
https://github.com/seek4science/seek/issues/2557


test/factories/extended_metadata_types.rb
```diff
@@ -16,11 +16,26 @@
     association :sample_attribute_type, factory: :string_sample_attribute_type
   end
 
-  factory(:datetime_extended_metadata_attribute,class:ExtendedMetadataAttribute) do
+  factory(:datetime_extended_metadata_attribute, class: ExtendedMetadataAttribute) do
     title { 'date' }
```
stuzart: would be clearer if this was `datetime` to match the type

lib/seek/rdf/rdf_generation.rb
```diff
@@ -179,6 +176,55 @@ def dependent_items
...
         queue_rdf_generation(true, false)
       end
+
+      private
+
+      def emit_emt_attributes(rdf_graph, subject, emt_type, data)
+        emt_type.extended_metadata_attributes.each do |attribute|
+          next unless attribute.pid.present?
+
+          value = data[attribute.accessor_name]
```
stuzart: should be accessed via `get_attribute_value` rather than the `data` hash directly.

lib/seek/rdf/rdf_generation.rb
```diff
@@ -179,6 +176,55 @@ def dependent_items
...
+          RDF::Literal(value.to_s, datatype: RDF::XSD.date)
+        when Seek::Samples::BaseType::DATE_TIME
+          RDF::Literal(value.to_s, datatype: RDF::XSD.dateTime)
+        when Seek::Samples::BaseType::INTEGER
+          RDF::Literal(value.to_i, datatype: RDF::XSD.integer)
+        when Seek::Samples::BaseType::FLOAT
+          RDF::Literal(value.to_f, datatype: RDF::XSD.double)
+        when Seek::Samples::BaseType::BOOLEAN
+          RDF::Literal(value, datatype: RDF::XSD.boolean)
+        else
```
stuzart, line 240: none of this would be necessary if the value was of the right object type. Would be better to cast to the right type according to the base type, and then simply use `RDF::Literal(value)`. The ruby rdf library handles the rest. For now could have a method here that casts it, but a better change would be do so through the base type handlers. Maybe either a flag to get_attribute_value - e.g `get_attribute_value(attr, cast:true|false)` or a separate method. This would make it consistent with the other rdf generation, the code cleaner and also fix things for samples

stuzart (changes requested):
there are still unresolved copilot comments that are valid.

long term I think we need a cleaner way of accessing the nested metadata without directly interacting with the hash, in a way that we can get it converted to the right type - either through the attribute, or through the main resource (e.g an attribute name that combines the keys - e.g `parent:child:name` ).  
the base_type_handler already validate and convert, and I think just need a method to cast to the correct type - but for this PR maybe just update the rdf_literal method to cast it.

PR #2569, opened 2026-05-07
Redis implementation for Session Store

kdp-cloud:
- Uses Redis as session store
- Docker compose update
- Ansible update


docker-compose.yml
```diff
@@ -34,12 +36,34 @@ services:
...
       timeout: 5s
       retries: 5
       start_period: 20s
 
+  session_store:
+    image: redis:8.6-alpine
+    container_name: seek-session-store
+    command: ["redis-server", "--appendonly", "yes"]
+    volumes:
+      - seek-session-data:/data
```
fbacall: I would change these names to something more generic (maybe just `redis`), since we will use it for more than just sessions.

.github/workflows/ansible-install.yml
```diff
@@ -26,6 +26,16 @@ jobs:
...
+      session-store:
+        image: redis:8.6-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 5
+
```
fbacall: The ansible deployment scripts (`script/ansible/Deploy-SEEK.yml`) need to be updated to actually install redis

fbacall (commented):
Looks good, but a couple of points:

- This will require all users to re-authenticate, unless we try and migrate sessions - should we?
- If not, then we should also drop the sessions table.

fbacall (approved):

Issue #2527, opened 2026-03-18
Consider mandatory attributes when detecting sample type / e. metadata type during FDS import


stuzart:
When importing, it is possible to detect the closing matching sample type (or EMT) which has mandatory attributes that are missing. This results in a validation and save failure. It should take this into account when detecting the closest match.

PR #2578, opened 2026-05-15
reject sample or extended metadata type candidates that have a required field that isn't present in the FDS import rdf

stuzart:
* fix for #2527

skips the types where there isn't a matching property value for the pid's for a required attribute.

### DIFF
--- a/lib/seek/fair_data_station/base.rb
+++ b/lib/seek/fair_data_station/base.rb
@@ -155,12 +155,13 @@ module Seek
 
       def find_closest_matching_extended_metadata_type(property_ids = additional_metadata_annotations.collect { |annotation| annotation[0] })
         candidates = ::ExtendedMetadataType.where(supported_type: type_name).enabled.includes(:extended_metadata_attributes).filter_map do |emt|
-          extended_metadata_property_ids = emt.deep_extended_metadata_attributes.collect(&:pid).compact_blank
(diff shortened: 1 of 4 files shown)


lib/seek/fair_data_station/base.rb
```diff
@@ -153,20 +153,18 @@ def find_exact_matching_extended_metadata_type
...
       end
 
-      def find_closest_matching_extended_metadata_type(property_ids = additional_metadata_annotations.collect do |annotation|
-        annotation[0]
-      end)
-        # collect and sort those with the most properties that match, eliminating any where no properties match
-        candidates = ::ExtendedMetadataType.where(supported_type: type_name).enabled.includes(:extended_metadata_attributes).collect do |emt|
+      def find_closest_matching_extended_metadata_type(property_ids = additional_metadata_annotations.collect { |annotation| annotation[0] })
+        candidates = ::ExtendedMetadataType.where(supported_type: type_name).enabled.includes(:extended_metadata_attributes).filter_map do |emt|
           extended_metadata_property_ids = emt.deep_extended_metadata_attributes.collect(&:pid).compact_blank
```
whomingbird: `emt.deep_extended_metadata_attributes ` has been called twice (also in line 163) maybe should cache the result.

whomingbird (changes requested):

whomingbird (approved):

Issue #2554, opened 2026-04-27
ObservationUnit not supported in ISA temporary sharing links


whomingbird:
Unlike other ISA assets, ObservationUnit is not handled in `auth_by_code?`, so temporary sharing links do not propagate access to ObservationUnits as they do for other items in the hierarchy.
Related to #758.

PR #2581, opened 2026-05-18
Allow temp links for ObservationUnit and include in auth chain for parent Study #2554

stuzart:
## Summary

- Adds temporary access link (special auth code) support for `ObservationUnit`, enabling the sharing link UI on its manage and show pages.
- Extends the auth code chain so that assets linked to an `ObservationUnit` (via `observation_unit` belongs_to or `observation_units` has_many) are granted access when a valid code exists on the `ObservationUnit`, its parent `Study`, or parent `Investigation`.
- Conditionally mentions observation units in the Study temp link help text when the feature is enabled.

Closes https://github.com/seek4science/seek/issues/2554

### DIFF
--- a/lib/seek/permissions/code_based_authorization.rb
+++ b/lib/seek/permissions/code_based_authorization.rb
@@ -22,18 +22,46 @@ module Seek
           return false
         elsif is_a?(Study) && respond_to?(:investigation)
           return investigation.special_auth_codes.unexpired.where(code: code).exists? if investigation
+        elsif is_a?(ObservationUnit)
+          if respond_to?(:study) && study
+            return true if study.special_auth_codes.unexpired.where(code: code).exists?
+            return study.investigation.special_auth_codes.unexpired.where(code: code).exists? if study.investigation
+          end
         elsif is_a?(Assay)
           if respond_to?(:study) && study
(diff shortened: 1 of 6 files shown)


lib/seek/permissions/code_based_authorization.rb
```diff
@@ -22,18 +22,46 @@ def auth_by_code?(code)
...
+              return true if ou.study.special_auth_codes.unexpired.where(code: code).exists?
+              return true if ou.study.investigation&.special_auth_codes&.unexpired&.where(code: code)&.exists?
+            end
+          end
 
-            if assay.study
-              return true if assay.study.special_auth_codes.unexpired.where(code: code).exists?
-              return true if assay.study.investigation&.special_auth_codes&.unexpired&.where(code: code)&.exists?
+          # Check via observation_units (plural has_many, e.g. DataFile via observation_unit_assets)
+          if respond_to?(:observation_units) && observation_units.any?
```
whomingbird: `observation_units.any?`  is this check redundant ? `observation_units.each `will do nothing if it is empty..

whomingbird (commented):
Just a nitpick. The rest is ok

PR #2586, opened 2026-05-19
ISA studies and ISA assays API endpoints

kdp-cloud:
- Make ISA studies and ISA assays controller methods available through API
- Add endpoints to API documentation


app/serializers/isa_study_serializer.rb
```diff
@@ -0,0 +1,86 @@
...
+      title: sample_type.title,
+      description: sample_type.description,
+      template_id: sample_type.template_id,
+      sample_attributes: serialize_isa_sample_attributes(sample_type),
+      samples: serialize_viewable_samples(sample_type)
+    }
+  end
+
+  def serialize_viewable_samples(sample_type)
+    sample_type.samples.select(&:can_view?).map do |sample|
```
whomingbird: sample_type.samples.select(&:can_view?) pulls all samples into memory and runs a per-record permission check, and the embedded array is never paginated. ISA sample types routinely hold thousands of samples, so a single show can be very expensive. Use the auth_lookup-backed scope instead, and consider capping the embedded set:                                                                  
` sample_type.samples.authorized_for('view').map do |sample|    `

app/serializers/isa_assay_serializer.rb
```diff
@@ -0,0 +1,98 @@
...
+      title: sample_type.title,
+      description: sample_type.description,
+      template_id: sample_type.template_id,
+      sample_attributes: serialize_isa_sample_attributes(sample_type),
+      samples: serialize_viewable_samples(sample_type)
+    }
+  end
+
+  def serialize_viewable_samples(sample_type)
+    sample_type.samples.select(&:can_view?).map do |sample|
```
whomingbird: the same concern as in `isa_study_serializer.rb`

whomingbird (changes requested):
Tested via UI and also with "curl",  it works

stuzart (approved):

PR #2590, opened 2026-05-20
Fix model files lost after validation error on create (#2582)

### LINKED ISSUE #2582: Model files lost after a validation error whilst creating a new entry
Only affects models, but due to the way the multiple file list is handled, if there is a validation error the files are still listed but the associated content blobs are lost. The mechanism for listing the files seems to have been intended for new versions but didn't account for handling validation errors.

stuzart:
## Summary

Fixes #2582 — uploaded model files were silently lost when a validation error occurred during creation and the user resubmitted the form.

**Root cause:** `ContentBlob` records built on an unsaved `Model` have no database ID. The `existing_file.hbs` template uses `{{id}}` to populate `retained_content_blob_ids[]` hidden fields, so with `id = nil` the fields submitted as empty strings and the files were discarded on resubmit.

- On validation failure, save the in-memory `ContentBlob` records as orphans (no `asset_id`) so they get real database IDs the form can reference
- On resubmit, load those orphaned blobs back into the in-memory association so they are treated as already-attached
- On successful save, update the orphaned blobs' `asset_id`/`asset_type`/`asset_version` to point to the newly created model
- Use the session as an ownership allowlist for retained blob IDs, preventing a user from attaching blobs they did not upload by tampering with params
- Fix client-side `validateUploadFormFields` in `validation.js` to also check `retained_content_blob_ids[]` inputs, preventing a spurious "Please specify at least a file/image" alert on resubmit
- Hide the "No file chosen" label beside the file input button when files are already shown in the pending list

Orphaned blobs from abandoned sessions are cleaned up by the existing `RegularMaintenanceJob#remove_dangling_content_blobs` after 8 hours.

## Changes

- `lib/seek/assets_standard_controller_actions.rb` — `preserve_content_blobs_for_rerender` and `attach_retained_orphaned_content_blobs`
- `lib/seek/upload_handling/data_upload.rb` — `load_orphaned_content_blobs` using `add_to_target`; use `safe_retained_content_blob_ids` throughout
- `lib/seek/upload_handling/parameter_handling.rb` — `retained_content_blob_ids` and `safe_retained_content_blob_ids` (session-based allowlist)
- `app/assets/javascripts/upload.js` — hide "No file chosen" text when files are already listed
- `app/assets/javascripts/validation.js` — check retained blob IDs in client-side validation; scope retained selector to parent when `parentId` is set
- `app/assets/stylesheets/styles.scss` — CSS to hide the file input label text when files are pending
- `test/functional/models_controller_test.rb` — end-to-end tests covering failed → successful resubmit and tampered ID rejection

Closes https://github.com/seek4science/seek/issues/2582

### DIFF
--- a/app/assets/javascripts/upload.js
+++ b/app/assets/javascripts/upload.js
@@ -36,7 +36,7 @@ $j(document).ready(function () {
         // Local file
         var addLocalFile = function () {
             var newField = HandlebarsTemplates['upload/file_field']();
-            $j(this).parent().append(newField);
+            $j(this).parent().append($j(newField).addClass('hide-no-file-text'));
             var filename = this.value.split(/\\/)[this.value.split(/\\/).length - 1];
             var listItem = $j(HandlebarsTemplates['upload/local_file']({ text: filename }));
             pending.append(listItem);
@@ -44,6 +44,12 @@ $j(document).ready(function () {
         };
 
         field.on('change', 'input[type=file][data-batch-upload=true]', addLocalFile);
+
+        // When existing files are already listed (e.g. after a validation error re-render),
+        // hide the "No file chosen" text beside the file input since files are already shown.
+        if (pending.children().length > 0) {
+            $j('input[type=file][data-batch-upload=true]', field).first().addClass('hide-no-file-text');
+        }
 
         // Remote file
         var addRemoteFile = function () {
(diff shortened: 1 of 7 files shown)


fbacall (commented):
Minor thing: when the form is loaded again after the validation error, you get a misleading flash error at the top of the page that reads:

>  Please select a file to upload or provide a URL to the data.

stuzart:
> Minor thing: when the form is loaded again after the validation error, you get a misleading flash error at the top of the page that reads:
> 
> > Please select a file to upload or provide a URL to the data.

@fbacall - updated now. The flash[:error] was being set within the `.none?` loop, so refactored to set the message outside of the loop instead of within `check_for_data_or_url`. Also fixed the message logic to only mention File and URL if data_url is present as a key, otherwise just ask for the file (previously was checking for data_url in both cases)

fbacall (approved):

PR #2597, opened 2026-05-28
Add optional flags to docker compose

kdp-cloud:
Optional flags were added to the command section of the docker compose, which might be useful when upgrading the MySQL DB.

### DIFF
--- a/docker-compose-relative-root.yml
+++ b/docker-compose-relative-root.yml
@@ -34,7 +34,7 @@ services:
       # Optional: Enable mysql_native_password if needed for legacy apps
       # - --mysql-native-password=ON
       # Optional: restrict foreign keys if needed
-      # - --restrict_fk_on_non_standard_key=OFF
+      # - --restrict-fk-on-non-standard-key=OFF
     restart: always
     stop_grace_period: 1m30s
     env_file:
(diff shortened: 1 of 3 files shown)


docker-compose-relative-root.yml
```diff
@@ -27,7 +27,14 @@ services:
...
     container_name: seek-mysql
-    command: --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --log-error-verbosity=1
+    command:
+      - --character-set-server=utf8mb4
+      - --collation-server=utf8mb4_unicode_ci
+      - --log-error-verbosity=1
+      # Optional: Enable mysql_native_password if needed for legacy apps
+      # - --mysql-native-password=ON
+      # Optional: restrict foreign keys if needed
+      # - --restrict_fk_on_non_standard_key=OFF
```
stuzart: Shouldn't this use hyphens instead of `_` -  ( https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_restrict_fk_on_non_standard_key ) . If either work, hyphens will match the other flags

If changed, the others files should be updated too.

stuzart (changes requested):

stuzart (approved):

PR #2610, opened 2026-06-09
Fix redis session store

kdp-cloud:
- Use `:redis_store` -> better integration wit rack
- Sets the `:secure` attribute of the red by default to false but lets you set it to true through env variables
- Let's you scope the session cookie to an FQDN if desired

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -6,8 +6,6 @@
 # FIXME: Use Seek::Config.session_store_timeout somehow
 
 session_url = "#{ENV.fetch('REDIS_URL', 'redis://localhost:6379/0')}/session"
-domain_name = ENV.fetch('SEEK_FQDN', 'localhost')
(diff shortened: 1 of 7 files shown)


config/initializers/session_store.rb
```diff
@@ -1,16 +1,20 @@
 # Be sure to restart your server when you modify this file.
 
-# Use the database for sessions instead of the cookie-based default,
+# Use Redis for sessions instead of the cookie-based default,
 # which shouldn't be used to store highly confidential information
 # (create the session table with "rails generate session_migration")
```
stuzart: I don't think this line is relevant anymore ?

docker-compose.yml
```diff
@@ -52,7 +53,7 @@ services:
       retries: 5
       start_period: 20s
 
-  redis:
+  redis_store:
```
stuzart, line 56: the other docker-composes will also need updating

stuzart (changes requested):

stuzart (approved):

Issue #952, opened 2022-04-06
Easier access to the main workflow file or link


stuzart:
When registering a workflow through github, or through manual upload, you can tag the main workflow file.

However, to access the file, you need to browse the files tab and find the download button for the file.

E.g. for https://workflowhub.eu/workflows/326 this is quite deep, and as a Jupyter notebook it would be convenient to easily get the link to copy into a Jupyter instance

PR #2615, opened 2026-06-10
Split download button

AbbyGavin:
Requested in #952 

image

The split button does not appear if the main workflow blob is nil.

### DIFF
--- a/app/views/assets/_asset_buttons.html.erb
+++ b/app/views/assets/_asset_buttons.html.erb
@@ -61,7 +61,6 @@
               <%= link_to image('run_galaxy') + ' ' + t('workflows.run.on_select'), '#', 'data-toggle' => 'modal', 'data-target' => '#run-galaxy-modal' -%>
             </li>
           <% end %>
-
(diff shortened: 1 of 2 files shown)


fbacall (commented):
Cheers, looks good! 

A few minor things:
- I think use `main_workflow_path` instead of `main_workflow_blob`, which will potentially perform some slow disk operations.
- For the "Download Main Workflow" link it should use a different icon - I think there is one just labelled `download` that we use for general download links.
- It could use a test to check it only displays the "Download Main Workflow" link if there is a main workflow (although there should always be a main workflow). Here's an example of a similar test to check a UI element: https://github.com/seek4science/seek/blob/34f70a528226bf0edd5ddc4203d522d42ad66b3f/test/functional/workflows_controller_test.rb#L210-L216

And a question - what does the "split" button look like if there is no main workflow?

AbbyGavin:
- Changed `main_workflow_blob.path` to `main_workflow_path`
- Using the download icon
- A new test will be added when I've fixed the ones that are currently failing
- The button just looks like a normal button if the split button isn't being displayed.

fbacall (approved):

PR #2642, opened 2026-06-26
Git `add_file` error handling

fbacall:
Fixes 500 error when trying to add a Git file without selecting a file to upload or providing a URL.

### DIFF
--- a/test/functional/git_controller_test.rb
+++ b/test/functional/git_controller_test.rb
@@ -71,6 +71,17 @@ class GitControllerTest < ActionController::TestCase
     refute assigns(:git_version).file_exists?('new-file.txt')
   end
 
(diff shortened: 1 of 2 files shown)


test/functional/git_controller_test.rb
```diff
@@ -71,6 +71,17 @@ def setup
     refute assigns(:git_version).file_exists?('new-file.txt')
   end
 
+  test 'displays error message if adding file without data or url' do
+    refute @git_version.file_exists?('little_file.txt')
```
stuzart: should this be checking for `new-file.txt` rather than little_file?

stuzart (changes requested):
just something very minor

stuzart (approved):

PR #2688, opened 2026-07-22
Thread-safety for `User.current_user` and `disable_authorization_checks`

fbacall:
Uses `Thread.current` to hold state instead of global/class vars.

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -226,7 +226,7 @@ class Sample < ApplicationRecord
 
   # checks and validates whether new linked samples have view permission, but ignores existing ones
   def validate_added_linked_sample_permissions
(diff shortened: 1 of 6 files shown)


stuzart (approved):
I think it looks safe. There are copilot comments, but I don't think we need to worry about fibres for Puma and Passenger - only fibre based servers. I'm not sure what copilot suggests are correct and would need looking into closer. If it's ever an issue `ActiveSupport::CurrentAttributes` and `IsolatedExecutionState` seem to be what needs looking into to handle it properly.

This fix alone doesn't necessarily mean jobs can run concurrently, there may be some the need to run in order. I can't think of any off the top of my head but I have a feeling there may be.

PR #2701, opened 2026-07-27
Not adding remote files

### LINKED ISSUE #1829: Unfetched remote files in Git result in empty file in RO-Crate
Should perhaps not write a file at all.

#1776

AbbyGavin:
Fixes #1829 

The directory that the crate is made from is created by a library, so we can't stop it being added in the first place. So we need to delete it from that folder before it is zipped into the RO-Crate.

This also adds the `localPath` property to remote files in downloaded crates. This is mainly because it is necessary for the deletion, but it is also allowed in the spec.


fbacall:
As discussed on Teams, handle the case where a `Git::Blob` is `remote?` but also `fetched?` (i.e. the user chose the "Fetch a copy" option when registering the file).

In these cases the file *should* be included in the RO-Crate zip (but the `@id` should still point to the URL, and `localPath` should still be set as with other remote files).

And make sure this is covered in the test.

fbacall (approved):

PR #2719, opened 2026-08-18
Readme relative links

### LINKED ISSUE #642: Relative links in workflow README.md used as description don't resolve
and result in 404s

Maybe we can display the README description in an iframe using the `/raw` paths?

AbbyGavin:
Fixes #642. This could be extended to other places that use the markdown renderer, but for now the description of workflows was the only pace that had reported theses issues.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -211,7 +211,7 @@ module ApplicationHelper
     return "<li><div class='none_text'> None specified</div></li>".html_safe if is_nil_or_empty?(list)
   end
 
-  def render_markdown(markdown, relative_root = '')
+  def render_markdown(markdown, relative_root: nil)
     Seek::Markdown.render(markdown, relative_root)
   end
 
@@ -230,7 +230,7 @@ module ApplicationHelper
       if options[:markdown]
         # Convert `&gt;` etc. back to `>` so markdown blockquotes can be used.
         # The markdown renderer will cope with rogue `>`s that are not part of quotes.
-        res = render_markdown(CGI::unescapeHTML(res), options[:markdown_relative_root])
+        res = render_markdown(CGI::unescapeHTML(res), relative_root: options[:markdown_relative_root])
       elsif options[:description] || options[:address]
         res = simple_format(res, {}, sanitize: false).html_safe
       end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -211,8 +211,8 @@ module ApplicationHelper
     return "<li><div class='none_text'> None specified</div></li>".html_safe if is_nil_or_empty?(list)
   end
 
-  def render_markdown(markdown)
-    Seek::Markdown.render(markdown)
+  def render_markdown(markdown, relative_root)
+    Seek::Markdown.render(markdown, relative_root)
   end
 
   def text_or_not_specified(text, options = {})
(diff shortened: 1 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -211,8 +211,8 @@ def empty_list_li_text(list)
     return "<li><div class='none_text'> None specified</div></li>".html_safe if is_nil_or_empty?(list)
   end
 
-  def render_markdown(markdown)
-    Seek::Markdown.render(markdown)
+  def render_markdown(markdown, relative_root = '')
```
fbacall: I think `relative_root` should be `nil` by default, and probably makes more sense to be a keyword argument (`relative_root: nil`)

app/helpers/application_helper.rb
```diff
@@ -230,7 +230,7 @@ def text_or_not_specified(text, options = {})
       if options[:markdown]
         # Convert `&gt;` etc. back to `>` so markdown blockquotes can be used.
         # The markdown renderer will cope with rogue `>`s that are not part of quotes.
-        res = render_markdown(CGI::unescapeHTML(res))
+        res = render_markdown(CGI::unescapeHTML(res), options[:markdown_relative_root])
```
fbacall: Pass as keyword argument

lib/seek/markdown.rb
```diff
@@ -19,14 +19,33 @@ def handle_element(a)
...
+
+      def selector
+        SELECTOR
+      end
+
+      def handle_element(img)
+        return if img['src'].nil? || context[:relative_root].nil?
+
+        src = img['src'].strip
+        return if src.start_with?('http')
```
fbacall: This check could erroneously pass if an image had a weird name like `http_my_website_com_logo.png` or something.

I think perhaps we should parse `img['src']` into a URI object and then check if it is `absolute?`.

test/functional/workflows_controller_test.rb
```diff
@@ -1765,6 +1765,26 @@ def bad_generator.write_graph(struct)
...
+    workflow = FactoryBot.create :ro_crate_git_workflow, policy: FactoryBot.create(:public_policy)
+
+    workflow.update(description: '<img src="test1/image.png"/><img src="/test2/image.png"/><img src="https://example.com/test3/image.png"/>')
+
+    get :show, params: { id: workflow }
+
+    assert_select '.seek-description', count: 1
+    assert_select '.seek-description img', count: 3
+    assert_select '.seek-description img:nth-of-type(1)' do
+      assert_select ":match('src', ?)", %r{/workflows/\d*/git/1/test1/image.png}
```
fbacall: Should these paths not be `/workflows/\d*/git/1/test1/raw/image.png` etc.?

fbacall (changes requested):
A couple of comments.

But a slightly bigger thing: I think maybe we should check for, and prevent path traversal. e.g. `images/../../../something.png`. I know we discussed it briefly and said it was probably not a huge risk for images, but I still think it's worth looking into a way to prevent it - there could be an easy method somewhere in `Addressable` to do it.

If we could prevent path traversal outside the base `/workflows/123/git/1/test1/raw/`, that would be great, for example: `docs/../images/123.png` would be OK, but `docs/../../images/456.png` wouldn't.

PS. if you want a more direct way of testing the markdown rendering, you could just call it directly and check the output as a string like in these tests: https://github.com/seek4science/seek/blob/8cebd91202bbb609cc32cd4a39e27bd543bac84d/test/unit/helpers/application_helper_test.rb#L426-L436

test/functional/workflows_controller_test.rb
AbbyGavin: They should be. One of the last things that I did before making these tests was to remove the `path` argument from the polymorphic path, which I didn't realise actually broke the paths. This is now fixed.

fbacall (approved):

PR #2721, opened 2026-08-20
Migrate background jobs from DelayedJob to Solid Queue (#2656)

### LINKED ISSUE #2656: Investigate using SolidQueue for managing the job queue
Replace the now dated delayed-job queue management with SolidQueue
https://guides.rubyonrails.org/active_job_basics.html#default-backend-solid-queue

stuzart:
Migrates SEEK's background job processing from DelayedJob to [Solid Queue](https://github.com/rails/solid_queue), and adds a [Mission Control – Jobs](https://github.com/rails/mission_control-jobs) dashboard for inspecting the queues. Closes #2656.

Solid Queue's tables live in the primary database (shared, not a separate queue DB). The `delayed_job_active_record` gem and its tables are retained so existing jobs can be migrated across (see the migrator below); they can be removed in 1.20.0.

## What changed

- **Adapter cutover** — `config.active_job.queue_adapter` flipped to `:solid_queue`; Solid Queue tables added via a regular migration (`db/migrate/20260715145804_create_solid_queue_tables.rb`).
- **Worker topology** (`config/queue.yml`) — one worker per queue, each `threads: 1` (a hard constraint while `User.current_user` / `$authorization_checks_disabled` remain process-global), feature-flag gated to match the previous per-queue setup. Started via `bin/jobs`; `script/run_solid_queue.sh` wraps it with restart-on-exit for deployment.
- **Recurring schedules** (`config/recurring.yml`) — replaces the `whenever`/`schedule.rb` cron generation. The `whenever` gem is removed; a static crontab (`docker/seek.crontab`) fed to supercronic now covers the one OS-level shell task (the soffice reaper). Schedules documented in-file with a cron-syntax reference.
- **Historical job migrator** — `seek:upgrade` migrates any leftover `delayed_jobs` rows into Solid Queue, preserving serialized arguments, folding `attempts` into `executions`, and tolerating dangling GlobalIDs.
- **Admin panels** — job queue stats, worker status, and the "restart background job workers" button reworked against Solid Queue's supervisor/process model.
- **Mission Control – Jobs dashboard** — mounted at `/jobs`, gated behind SEEK's own admin auth (not the gem's HTTP Basic auth). Finished jobs retained for 14 days.
- **rake tasks** — `jobs:work`/`workoff`/`clear`/`check` redefined against Solid Queue.

## Behaviour changes worth noting

- The weekly subscription-digest schedule moved to `0 0 * * 0` (every Sunday). The old `whenever`-generated `0 0 1,8,15,22 * *` left a coverage gap near month boundaries against the job's fixed `1.week.ago` window — a long-standing bug, not a regression, now fixed.
- `db:sessions:batch_trim` dropped (sessions live in Redis with `allkeys-lru` now).

## Testing

Covered by `test/integration/recurring_test.rb`, `test/unit/delayed_job_migrator_test.rb`, updated admin/status tests, and new unit/functional coverage for the supervisor-pid and worker-restart edge cases.

## Draft — still open

- The `delayed_job_active_record` gem and its tables are retained so existing DelayedJob rows can be migrated into Solid Queue; they can be removed in 1.20.0 once no instance still needs to migrate.
- Kept as a documented non-issue: command-based recurring tasks serialize on the single `solid_queue_recurring` worker thread (only effect is a few minutes of overnight status-cache staleness while the daily bioschema dump runs).


fbacall (commented):
The `seek:workers` tasks were removed, but we have various init scripts that call out to those that would need updating.

Wondering if it would be worth keeping them to make updating simpler (and to abstract away the job backend)

stuzart:
> The `seek:workers` tasks were removed, but we have various init scripts that call out to those that would need updating.
> 
> Wondering if it would be worth keeping them to make updating simpler (and to abstract away the job backend)

Fair point. I did recreate the `jobs:` tasks for this reason, and don't see why not for `seek:workers`. I think they can just delegate to the jobs tasks

stuzart:
> The `seek:workers` tasks were removed, but we have various init scripts that call out to those that would need updating.
> 
> Wondering if it would be worth keeping them to make updating simpler (and to abstract away the job backend)

@fbacall - I've changed this now. Required a few more changes as it replaces a shell script that was used in various changes, and required a change in how the restart button works

fbacall (approved):

test/integration/recurring_test.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    sitemap = pop_task(:sitemap_refresh)
+    assert_equal 'SitemapRefreshJob', sitemap[:class]
+    assert_equal '45 0 * * *', sitemap[:schedule]
+
+    assert_empty @tasks, "Found untested recurring task(s): #{@tasks.keys.join(', ')}"
+  end
+
+  test 'news feed refresh schedule changes with config' do
+    with_config_value(:home_feeds_cache_timeout, 731) do
```
kdp-cloud: Just pointing out that this is not a valid task anymore. SolidQueue, which uses Fugit for Cron validation, crashes the background processes if the frequency is not a number between 1 and 99.
I don't know if it's worth looking into validating the value coming from the config? Or maybe document this?

kdp-cloud (changes requested):
LGTM in general.
There is just a test that looks a bit odd to me.

test/integration/recurring_test.rb
stuzart: thanks, I'll take a look

stuzart: I've applied a fix.

kdp-cloud (approved):

PR #2722, opened 2026-08-21
Handle non UTF-8 and very large content when rendering text previews

stuzart:
Viewing an asset whose content blob is typed as text (e.g. `text/plain`, `text/xml`) but is not valid UTF-8 raised `ArgumentError: invalid byte sequence in UTF-8` from `app/helpers/assets_helper.rb`, so the show page could not be rendered at all.

Content read from a blob is tagged as UTF-8 without being validated. `CGI.escapeHTML` (used by `h`) passes invalid bytes straight through, so the invalid string escaped the renderer's own rescue and raised when the helper called `String#blank?` on it, since that runs a regexp match. `TextRenderer` now replaces invalid bytes, as `Seek::Markdown.render` already does.

Rendering a preview also read the whole file into memory, escaped it and cached the result, which is expensive for a large file and repeats for every viewer. Content is now limited to `TextRenderer::MAX_RENDERABLE_SIZE` (1MB), with a message indicating that it is truncated. The limit matches `cache_max_redis_item_size`, so a cached preview stays within a single Redis item.

Both the inline preview and the standalone view are affected.

### DIFF
--- a/lib/seek/renderers/blob_renderer.rb
+++ b/lib/seek/renderers/blob_renderer.rb
@@ -3,6 +3,10 @@ module Seek
     class BlobRenderer
       include ActionView::Helpers
 
+      # forms part of the key for cached rendered content, and is bumped when a change to the
+      # renderers should invalidate content that has already been cached
+      CACHE_VERSION = 2
+
       attr_reader :blob, :params
 
       def initialize(git_blob_or_blob, url_options: {}, params: {})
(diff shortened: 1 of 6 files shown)


lib/seek/renderers/blob_renderer.rb
```diff
@@ -3,6 +3,10 @@ module Renderers
     class BlobRenderer
       include ActionView::Helpers
 
+      # forms part of the key for cached rendered content, and is bumped when a change to the
+      # renderers should invalidate content that has already been cached
+      CACHE_VERSION = 2
```
fbacall: Do we need this? I feel like the same issue affects all caches. Should we just clear the cache during upgrades?

fbacall (approved):

